PR #8, opened 2021-06-30
Split kinds of experts out of the constituencies list.

jyasskin:
I didn't attempt to put the lists in any sort of order, except to put the chairs next to the WHATWG SG.


torgo (approved):
lgtm

Issue #1, opened 2021-06-15
Figure out the default privacy boundary for the web


jyasskin:
The [origin](https://html.spec.whatwg.org/multipage/origin.html#origin) is the default _security_ boundary for the web. That is, different things within the same origin are expected to be able to interfere with each other, while cross-origin things have to opt into communication. There are some exceptions, like cookies and [Chrome's site isolation](http://www.chromium.org/Home/chromium-security/site-isolation#TOC-Current-Status), but we have consensus that origin isolation is the goal.

A default privacy boundary would, roughly, be the point at which the user (rather than the page) needs to approve communication.

We're currently defaulting to the [Public Suffix List](https://publicsuffix.org/) for this, but that answer isn't working perfectly. Specifically, the PSL groups origins into sites, and we declare that the site is the privacy boundary. The PSL has known [problems](https://github.com/sleevi/psl-problems), and it's not currently funded well enough to handle wider adoption, especially in [potentially-adversarial situations](https://github.com/privacycg/private-click-measurement/issues/78). We might be able to round up the needed extra investment if we so choose.

A [variant of First Party Sets](https://github.com/privacycg/first-party-sets#origins-instead-of-registrable-domains) might be able to serve as an alternate default privacy boundary.

jyasskin:
I forgot to mention that the IETF has tried to address an adjacent problem, but didn't succeed: https://datatracker.ietf.org/wg/dbound/about/ and https://github.com/equalsJeffH/dbound#readme.

torgo:
As discussed today - we think building on the [whatwg def](https://html.spec.whatwg.org/multipage/origin.html#origin) of origin and writing some words around that might be the best approach here. We could write it from the perspective of how new technologies should behave.

jyasskin:
[...]

PR #28, opened 2021-08-04
Rewrite the identity section to try to define the default enforceable privacy/contextual boundary.

jyasskin:
This improves #1 .

This incorporates some ideas from https://github.com/asankah/identity-domains,
which distinguishes separate profiles and boundaries a user creates by clearing
cookies/storage. This covers one of the things @sandandsnow wanted to be sure to cover, but I think I forgot a second thing.

It explicitly says that browsers are free to separate contexts more finely than
this default, which I hope covers @pes10k's concern in this area. We should add a general statement about browsers being free to more-aggressively protect their users in a separate PR.

It drops the blanket statement that automating recognition is always
inappropriate, as all webservers are automated, and sometimes users intend to
present the same identity in multiple contexts.

It also removes the explicit mention of email-based cross-context recognition
in favor of a more general statement about difficult-to-forge pieces of
people's identities.

Let me know what you think.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/28.html#identity" title="Last updated on Aug 18, 2021, 9:29 PM UTC (d3668c3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/28/a5107a0...jyasskin:d3668c3.html" title="Last updated on Aug 18, 2021, 9:29 PM UTC (d3668c3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -535,7 +535,7 @@ their agency and sovereignty.
 A [=person=]'s <dfn>identity</dfn> is the set of characteristics that define them. Their identity
 *in a [=context=]* is the set of characteristics they present to that context. People
 frequently present different identities to different contexts, and also frequently share an identity
-between several contexts.
+among several contexts.
 
 <dfn>Cross-context recognition</dfn> is the act of recognising that an [=identity=] in one
 [=context=] is the same [=person=] as an [=identity=] in another [=context=]. [=Cross-context
--- a/index.html
+++ b/index.html
@@ -570,11 +570,18 @@ default, [=user agents=] define a <dfn>machine-enforceable context</dfn> or <dfn
 * between points in time that the user or their agent clears that [=site=]'s cookies and other
   storage (which is sometimes automatic at the end of each session).
 
-Even though this is the default, [=user agents=] are free to both widen and restrict this context as
+Even though this is the default, [=user agents=] are free to restrict this context as
 their users need. For example, some user agents may help their users present different
-[=identities=] to subdivisions of a single [=site=], while others may help their users present a
-single [=identity=] to a [=site=] across multiple installations or to multiple [=sites=] that the
-user understands represent a single [=party=].
+[=identities=] to subdivisions of a single [=site=].
+
+<div class="issue" data-number="1">
+
+There is disagreement about whether [=user agents=] may also widen their [=machine-enforceable
+contexts=]. For example, some user agents might want to help their users present a single
+[=identity=] to multiple [=sites=] that the user understands represent a single [=party=], or to a
+[=site=] across multiple installations.
+
+</div>
 
 [=User agents=] should prevent their user from being [=cross-context recognition|recognized=] across
 [=machine-enforceable contexts=] unless the user intends to be recognized. This is a "should" rather
--- a/index.html
+++ b/index.html
@@ -223,7 +223,7 @@
     };
   </script>
 </head>
-<body data-cite="html indexedDB service-workers fingerprinting-guidance">
+<body data-cite="html indexedDB service-workers fingerprinting-guidance url">
 
 <section id="abstract">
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -536,34 +536,54 @@
...
+In computer systems and on the Web, an [=identity=] seen by a particular website is typically
+assigned an <dfn>identifier</dfn> of some type, which makes it easier for an automated system to
+store data about that user.
+
+[=User agents=] support their [=users=]' [=autonomy=] by helping them present their intended
+[=identity=] to each [=context=] that they visit. To do this, [=user agents=] have to make some
+assumptions about the borders between [=contexts=]. By default, [=user agents=] define a
+<dfn>machine-enforceable context</dfn> or <dfn>partition</dfn> as:
+
+* A set of [=environments=] (roughly top-level pages, iframes, and workers)
```
jyasskin, line 562: Reorder to emphasize that the iframes and workers don't have to be top-level:

```suggestion
* A set of [=environments=] (roughly iframes (including third-party iframes), workers, and top-level pages,)
```

index.html
```diff
@@ -536,34 +536,54 @@
...
-A keystone principle of the Web is trust [[RFC8890]]. An important part of trust is to
-ensure that [=data=] collected for a [=purpose=] that matches a clearly delineated [=user=]
-feature should not then be used for additional secondary [=purposes=]. Email is often used
-for login and communication [=purposes=], including essential transactional interactions.
-Using emails for cross-context recognition [=purposes=] is therefore not only
-[=inappropriate=] but also threatens key messaging infrastructure. Future iterations of
-the Web platform should ensure that users can log into sites and receive communication
-without relying on email at all.
+A [=person=]'s <dfn>identity</dfn> is the set of characteristics that define them. Their identity
+*in a [=context=]* is the set of characteristics they choose to present to that context. People
```
jyasskin, line 540:
```suggestion
*in a [=context=]* is the set of characteristics they present to that context. People
```

index.html
```diff
@@ -536,34 +536,54 @@
...
+you expect them to have rather different discussion topics with you than they usually would, and
+possibly even to pretend they do not know you.) This is particularly true for [=vulnerable=] people
+as recognising them in different [=contexts=] may force their vulnerability into the open.
+
+In computer systems and on the Web, an [=identity=] seen by a particular website is typically
+assigned an <dfn>identifier</dfn> of some type, which makes it easier for an automated system to
+store data about that user.
+
+[=User agents=] support their [=users=]' [=autonomy=] by helping them present their intended
+[=identity=] to each [=context=] that they visit. To do this, [=user agents=] have to make some
```
jyasskin, line 558: Highlight this as a ***principle***. Copy formatting ideas from https://www.w3.org/TR/fingerprinting-guidance/#detectability.

index.html
```diff
@@ -536,34 +536,54 @@
...
+* being visited within the same user agent installation (and profile for user agents that support
+  multiple profiles)
+* between points in time that the user clears that [=site=]'s cookies and other storage
+  (which some user agents do automatically at the end of each session).
+
+Even though this is the default, [=user agents=] are free to both widen and restrict this context as
+their users need. For example, some user agents may help their users present different
+[=identities=] to subdivisions of a single [=site=], while others may help their users present a
+single [=identity=] to a [=site=] across multiple installations or to multiple [=sites=] that the
+user understands represent a single [=party=].
```
jyasskin, line 574: We could draw lines around what variation is acceptable, but it's not clear to me what those lines should be, so it'll wait for a future PR.

index.html
jyasskin: Fingerprinting Guidance uses https://respec.org/docs/#practice. I've used that, and we can put a `<section id="bp-summary">` somewhere to get an index for free, but it says "Best Practices" instead of "Principles".

index.html

index.html
jyasskin: Tweaked.

index.html
```diff
@@ -536,34 +536,61 @@
...
+* A set of [=environments=] (roughly iframes (including cross-site iframes), workers, and top-level
+  pages)
+* whose [=environment/top-level origins=] are in the [=same site=] (roughly, their [=origin/hosts=]
+  have the same [=host/registrable domain=], but see [[PSL-PROBLEMS]])
+* being visited within the same user agent installation (and profile for user agents that support
+  multiple profiles)
+* between points in time that the user or their agent clears that [=site=]'s cookies and other
+  storage (which is sometimes automatic at the end of each session).
+
+Even though this is the default, [=user agents=] are free to both widen and restrict this context as
```
jonathanKingston, line 577: Should this text be part of the document? First party sets are still pretty well contested across other vendors; I'm not sure we should acknowledge them.

pes10k: I agree with this, and @jonathanKingston has put more directly and concisely what I tried to mention on the call:

If there are context boundaries that are controversial, or might be ruled out by the principals mentioned elsewhere in the document, I dont think we should include those context boundaries in a menu of examples.

darobin: I agree that FPS are still heavily contested, however I don't want to just drop things that are contentious. We should be using documents like this one to hash things out.

This seems like a good place to flag an in-document issue, no?

```
☣️⚠️☢️ WARNING ☣️⚠️☢️
People disagree on boundaries yada yada…
```

jyasskin: There are several examples of user-serving identity partitions that are excluded by the purely-site-based definition above here. I understand that Mozilla's currently shipping [one of them](https://github.com/disconnectme/disconnect-tracking-protection/blob/master/entities.json), even though it would like to get rid of that in the long run. Others include:

* bbc.com/bbc.co.uk
* github.com/githubusercontent.com and google.com/googleusercontent.com, where the domains are separate for security reasons, but users still want a uniform identity across the pair in order to access ACL'ed content
* news.bbc/sports.bbc where the registrable-domain algorithm doesn't understand that a whole TLD can be a very expensive domain registration

My text doesn't endorse any particular solution to the problem, and it restricts solutions to ones where "the user understands [the sites] represent a single party", but it does say that it's reasonable for browsers to explore solutions. 

I'm happy to mark this as contentious if y'all still think the document should forbid browsers from working on this problem as a whole.

This paragraph also gives the example of users sync'ing their identity<->site mapping between devices. That can't be done by naively sync'ing all cookie values, but is that kind of goal also contentious?

darobin: The idea is not to forbid browsers from working on this (at all), but on the contrary to mark this as an area that needs significant further discussion. We do want to get to a resolution before the document is finalised, though.

jonathanKingston: Not to get too weedsy on the discussion regarding FPS but Mozilla's implementation is [off by default](https://github.com/w3ctag/design-reviews/issues/342#issuecomment-820551099) purely a web compat solution and they [don't wish to see it standardised](https://github.com/privacycg/proposals/issues/17#issuecomment-641691610).

Either way I'm happy to just mark this as "in discussion" and we should resolve before publishing.

jyasskin: I've moved the mention of widening partitions into a "This is controversial" issue block. How's it look now?

index.html
```diff
@@ -532,34 +532,67 @@
...
-feature should not then be used for additional secondary [=purposes=]. Email is often used
-for login and communication [=purposes=], including essential transactional interactions.
-Using emails for cross-context recognition [=purposes=] is therefore not only
-[=inappropriate=] but also threatens key messaging infrastructure. Future iterations of
-the Web platform should ensure that users can log into sites and receive communication
-without relying on email at all.
+A [=person=]'s <dfn>identity</dfn> is the set of characteristics that define them. Their identity
+*in a [=context=]* is the set of characteristics they present to that context. People
+frequently present different identities to different contexts, and also frequently share an identity
+between several contexts.
```
wseltzer, line 538: s/between/among/

jyasskin: Done, thanks.

index.html
jyasskin: Discussion at the [Aug 18 meeting](https://github.com/w3ctag/privacy-principles/blob/main/meetings/2021-08-18-minutes.md#identity-and-boundary) indicated this was good enough for now. I'll merge tomorrow to give a bit of time for more asynchronous comments.

PR #32, opened 2021-08-05
Can we define party by accountability, not ownership?

dmarti:
It is possible for the same "company" that is ultimately owned by the same stockholders to be more than one party for purposes of being accountable for user data.

There are situations where the same company's web properties in different countries do have common ownership, but users in one jurisdiction are party to a EULA with one subsidiary and users in another jurisdiction have a EULA with a different entity.

From a user point of view the top level owner does not matter as much as whatever legal requirements the "party" is subject to for how their data is used.

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/pull/32.html" title="Last updated on Aug 5, 2021, 7:33 PM UTC (b5d6323)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/32/a1c497f...b5d6323.html" title="Last updated on Aug 5, 2021, 7:33 PM UTC (b5d6323)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -346,7 +346,7 @@ intellectual or psychological impairment, or refugees.
 ## The Parties {#parties}
 
 A <dfn>party</dfn> is a [=person=], a legal entity, or a set of legal entities that
-share common owners, common controllers, and a group identity that is readily evident to
+share legal accountability for use of data, common controllers, and a group identity that is readily evident to
 the [=user=] without them needing to consult additional material, typically through
 common branding.
 


sandandsnow:
For discussion: While I like the idea behind this suggestion, i.e. to move away from ownership of companies, closer to the concept of a data controller (via accountability), which is a commonly used term in privacy and data protection frameworks. However, I still see problems with extending the definition of first party to a set of legal entities. Please remind me, what is the reasoning behind having a definition of first party that includes more than one legal entity?

jyasskin:
We discussed this on [Aug 4](https://github.com/w3ctag/privacy-principles/blob/main/meetings/2021-08-04-minutes.md#:~:text=a%20set%20of-,legal,-entities%3F%20Isn%27t%20a), without @dmarti present, and I think we concluded that we could simplify to having a single legal entity. We lost that conclusion in subsequent discussions, but it's probably still correct.

The move to "accountability" is probably right in text that talks about multiple parties, especially multiple first-parties. I see [1 example](https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/pull/32.html#dfn-first-parties) of that in the current document?

Issue #34, opened 2021-08-17
Re: 3.1 Unwanted same-site recognition


sandandsnow:
* Propose amending the sentence “Users of most instantiations of the web platform expect that if they visit a site on one day, and then visit again the next day, the site will be able to recognize that they're the same user” to add words along the lines of “but they generally do not expect that the site will track their interaction with the site over time (such as everything they have viewed, for how long, etc.)”. Propose replacing “this” in the following sentence to “Same-site recognition allows …”.
* Would it make sense to have a separate section entitled “Unwanted same-site tracking”?
* Editorial: Remove “only” from “A privacy harm only occurs …”
* In which category would we describe the privacy harm from a single instance of site-recognition? (e.g., where a site is inappropriately detecting data to infer whether a user has a visual or other impairment).
* Consider adding synonymous terms to “private browsing” such as incognito mode. Also, add effectively synonymous terms for “profiles” such as containers or container tabs.
* Editorial: Add “or more” after “expert their two visits ...”
* Add to the list where a user might expect their two visits will not be associated:
	* Use extensions or add-ons to block tracking
	* Where they believe privacy or data protection laws would not permit tracking
* Even though this is discussed in the opening text of the section, considering adding a sentence “A user might expect that their two visits would be associated for specific purposes such as to maintain the contents of their shopping cart or to avoid having to log in again”.
* For discussion: What recommendations (namely, MUST or SHOULD statements) should we include for this section? A starting point regarding fingerprinting appears in the PING Group Note on [Mitigating Browser Fingerprinting in Web Specifications]( https://www.w3.org/TR/fingerprinting-guidance/).

darobin:
@sandandsnow to double check that this is in and close if so.

sandandsnow:
Re: 2.1.1.1. Same-site recognition, suggest amending these parts of the text as follows:
" ... While [User agents](https://w3ctag.github.io/privacy-principles/#dfn-user-agent), in general, are not able to determine exactly where intra-site [context](https://w3ctag.github.io/privacy-principles/#dfn-context) boundaries are, or how a site allows a [person](https://w3ctag.github.io/privacy-principles/#dfn-individual) to express that they intend to change [identities](https://w3ctag.github.io/privacy-principles/#dfn-identity) and/or they do not wish site to recognise them on subsequent visits, they are not able to enforce that sites actually separate [identities](https://w3ctag.github.io/privacy-principles/#dfn-identity) at those boundaries. 

[...]

PR #38, opened 2021-08-24
add initial draft of section on unwanted profiling, fixes #27

pes10k:
Adds initial draft of new "Privacy principles by category", describing unwanted profiling, or the privacy harm of sites learning information about a person, the person doesn't wish to share, even if that information cannot be used for fingerprinting and isn't "sensitive"

PR #40, opened 2021-08-26
Refine the cross-context recognition threats now that the #identity section defines the concept.

jyasskin:
This tries to fix half of #34.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/40.html#hl-recognition-cross-context" title="Last updated on Oct 6, 2021, 4:29 PM UTC (4d02e3b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/40/b01f37b...jyasskin:4d02e3b.html" title="Last updated on Oct 6, 2021, 4:29 PM UTC (4d02e3b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -925,7 +925,7 @@ technical feature that is fundamental to the Web.</span></p>
 * This principle uses "quickly" because it currently appears impossible to prevent some forms of
   fingerprinting that take a long time or many visits within each [=partition=].
 * This principle is limited in cases that only affect a small fraction of people who use the web
-  because people need to be allowed to configure their systems in unique ways, for example by using
+  because people may configure their systems in unique ways, for example by using
   a browser with a very small number of users. As long as a tracker can't track a significant number
   of people, it's likely to be unviable to maintain the tracker. However, this doesn't excuse making
   small groups of people trackable when those people didn't choose to be in the group.
@@ -969,7 +969,7 @@ Guidance § Clearing all local state</a> discusses how specifications can help
 browsers avoid this mistake.
 
 Fingerprinting consists of using attributes of the user's browser and platform
-that are consistent between two or more visits and probabilistically unique to the
+that are consistent between two or more visits and probably unique to the
 user.
 
 The attributes can be exposed as information about the user's device that is
--- a/index.html
+++ b/index.html
@@ -569,8 +569,8 @@ default, [=user agents=] define a <dfn>machine-enforceable context</dfn> or <dfn
 * A set of [=environments=] (roughly iframes (including cross-site iframes), workers, and top-level
   pages)
 * whose [=environment/top-level origins=] are in the [=same site=] (but see [[PSL-PROBLEMS]])
-* being visited within the same user agent installation (and browser profile for user agents that
-  support multiple browser profiles)
+* being visited within the same user agent installation (and browser profile, container, or
+  container tab for user agents that support those features)
 * between points in time that the user or their agent clears that [=site=]'s cookies and other
   storage (which is sometimes automatic at the end of each session).
 
@@ -590,9 +590,9 @@ contexts=]. For example, some user agents might want to help their users present
 [=User agents=] should prevent their user from being [=cross-context recognition|recognized=] across
 [=machine-enforceable contexts=] unless the user intends to be recognized. This is a "should" rather
 than a "must" because there are many cases where the user agent isn't powerful enough to prevent
-recognition. For example if two services that a user needs to use insist that the user share a
-difficult-to-forge piece of their identity in order to use the services, it's the services behaving
-[=inappropriately=] rather than the [=user agent=].
+recognition. For example if two or more services that a user needs to use insist that the user share
+a difficult-to-forge piece of their identity in order to use the services, it's the services
+behaving [=inappropriately=] rather than the [=user agent=].
 
 If a [=site=] includes multiple [=contexts=] whose [=norms=] indicate that it's [=inappropriate=] to
 share data between the contexts, the fact that those distinct [=contexts=] fall inside a single
@@ -902,32 +902,28 @@ data-cite="RFC6973#section-5.2.5">RFC6973§5.2.5</a>.
 These threats combine into the particular concrete threats we want web
 specifications to defend against, described in subsections here:
 
-## Unwanted same-site recognition {#hl-recognition-same-site}
+## Unwanted cross-context recognition {#hl-recognition-cross-context}
 
 Contributes to [=surveillance=], [=correlation=], and [=identification=].
 
-Users of most instantiations of the web platform expect that if they visit a
-site on one day, and then visit again the next day, the site will be able to
-recognize that they're the same user. This allows sites to save the user's
-preferences, shopping carts, etc. The web platform offers many mechanisms that
-are either intended to accomplish this recognition or that can be trivially used
-for it, including [[RFC6265|cookies]], {{WindowLocalStorage/localStorage}},
-{{WindowOrWorkerGlobalScope/indexedDB}}, {{CacheStorage}}, and other forms of
-storage.
+As described in [[[#identity]]], [=cross-context recognition=] can sometimes be [=appropriate=], but
+web specifications should ensure users have as much control as possible over when websites do it
+between different [=partitions=]. There are two broad classes of recognition that specifications
+address in different ways.
 
-A privacy harm only occurs if the user wants to break the association between
-two visits, but the site can still determine with high probability that the two
-visits came from the same user.
+### Same-site recognition {#hl-recognition-same-site}
 
-A user might expect that their two visits won't be associated if they:
+The web platform offers many ways for a website to recognize that a user is using the same identity
+over time, including [[RFC6265|cookies]], {{WindowLocalStorage/localStorage}},
+{{WindowOrWorkerGlobalScope/indexedDB}}, {{CacheStorage}}, and other forms of storage. This allows
+sites to save the user's preferences, shopping carts, etc., and users have come to expect this
+behavior.
 
-* Use a browser that promises to avoid such correlation.
-* Use their browser's private browsing mode. ([[WHAT-DOES-PRIVATE-BROWSING-DO]])
-* Use two different browser profiles between the two visits.
-* Explicitly clear the site's cookies or storage.
+A privacy harm occurs if the user changes [=partitions=], for example by clearing cookies or
+switching browser profiles, but the site can still determine with high probability that the two or
+more visits came from the same user.
 
-This recognition is generally accomplished by either "supercookies" or <a
-data-cite="fingerprinting-guidance#dfn-browser-fingerprinting">browser
+This recognition is generally accomplished by either "supercookies" or <a>browser
 fingerprinting</a>.
 
 Supercookies occur when a browser stores data for a site but makes that data
@@ -937,7 +933,7 @@ Guidance § Clearing all local state</a> discusses how specifications can help
 browsers avoid this mistake.
 
 Fingerprinting consists of using attributes of the user's browser and platform
-that are consistent between the two visits and probabilistically unique to the
+that are consistent between two or more visits and probabilistically unique to the
 user.
 
 The attributes can be exposed as information about the user's device that is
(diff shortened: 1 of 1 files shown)


sandandsnow (commented):
Thank you for updating the draft @jyasskin. 

On same-site recognition, there may still be privacy harm even if the use does not change partitions, for example if the user's intent was to in effect create partitions for each visit such as by using a browser that offers tracking protection, or extensions or add-ons to block tracking, choose no, deny or disallow tracking cookies when offered a choice, or maybe even a VPN to hide their location.

As the Privacy Threat Model is a work in progress, it may be preferable for now to leave out the dependency and instead to state the principles to be applied to this scenario.

Note: assuming this PR does not deal with unexpected same-site profiling (or whatever term we land on for that).

jyasskin:
> On same-site recognition, there may still be privacy harm even if the use does not change partitions, for example if the user's intent was to in effect create partitions for each visit such as by using a browser that offers tracking protection, or extensions or add-ons to block tracking, choose no, deny or disallow tracking cookies when offered a choice, or maybe even a VPN to hide their location.

Indeed. It seems like we have 2 kinds of harm here: harms in general, which in this section would happen if "the user changes **_contexts_** but the site **_does_** still determine that the two or more visits came from the same user." Versus harms that the UA should try to prevent even in the face of hostile sites, which is the text I wrote. Do we know of any existing terms that would distinguish those? Otherwise, I can make something up.

> As the Privacy Threat Model is a work in progress, it may be preferable for now to leave out the dependency and instead to state the principles to be applied to this scenario.

I don't have a good idea for principles that would draw a line between the harms we can prevent vs the harms-we-can't-prevent-without-breaking-the-web. In the meantime, I could turn this sentence into an issue paragraph that encourages us to figure out the relevant princples.

> Note: assuming this PR does not deal with unexpected same-site profiling (or whatever term we land on for that).

I'm trying to leave same-site profiling to #38.

jyasskin:
Here's my attempt to turn our [Sept 1 discussion](https://github.com/w3ctag/privacy-principles/blob/main/meetings/2021-09-01-minutes.md#pr-refine-the-cross-context-recognition-threats-now-that-the-identity-section-defines-the-concept) into text. I didn't do the stuff around "cookies and storage" in this PR because that phrase isn't in this part of the document.

jyasskin:
I rewrote this again, to unify the two cross-site recognition principles and to add a phrase that makes niche browsers ok; sorry if you'd started reviewing the last version.

sandandsnow:
@jyasskin, thank you very much for the re-writes of 3.1. Hopefully, I was reading the latest version when I prepared these comments today: 
- Editorial: change “a feature that is fundamental to the web” to “a feature that is fundamental to the architecture of the web"
- Editorial: change "because it appears impossible" to "because it currently appears impossible"
- Discuss: concern about allowing violations that only affect a small fraction of people because that small fraction could be vulnerable and, in any case, why should a small population experience a lower level of privacy on the web? Or am I missing something in that statement?
- Editorial: change “fundamental aspects of the web" to “fundamental architectural features of the web such as interoperability”
- Discuss: could we avoid using the “trade-off” terminology?
- Re: “and users have come to expect this behavior,” add "in such contexts" because user's would not always have that expectation.
- Re: "When a violation occurs at their divisions between different browser profiles or at the point a user clears their cookies and site storage .." consider adding other signals such as use of extensions or add-ons to blocking tracking and private or incognito mode.
- Re: "The web platform offers many ways for a website to recognize that a user is using the same identity over time .." - does this include ways for a website to recognise that a user has changed identity but is the same user?
- Discuss: "A privacy harm occurs if the user expresses that they intend ..". I understand the point of trying to tie the privacy harm to something that has been positively communicated by the user, but a privacy harm can occur if what happens does not match a user's expectation (whether or not they actively expressed their intent). Perhaps it would be helpful if I had a better understanding of what is meant by "if the user expresses that they intend ..".
- Discuss: user agent responsibility - (and putting that aside for the moment, I will make the observation that at least some browsers (if not all major browsers) offer users' at least some level of protection against same-site recognition.
- Editorial: change "personally-identifying information" to "personal data" or "personal information"
- Please remind me: where are we articulating the privacy harm of profiling that can occur even on a single visit to a site?

jyasskin:
Thanks for the review and for giving us lots of stuff to talk about tomorrow.

> * Editorial: change “a feature that is fundamental to the web” to “a feature that is fundamental to the architecture of the web"

I don't feel very strongly about this, but I think we ought to allow arguments about whether non-architectural features are fundamental to the web, as a defense of keeping some privacy problems. I don't have a concrete example, but if we somehow fixed a privacy problem without changing the web's architecture, but doing so drove all the users away to another platform, I'd like these principles to accept that as a mistake.

> * Editorial: change "because it appears impossible" to "because it currently appears impossible"

Changed.

> * Discuss: concern about allowing violations that only affect a small fraction of people because that small fraction could be vulnerable and, in any case, why should a small population experience a lower level of privacy on the web? Or am I missing something in that statement?

I added the "involuntary" bit to try to cover part of this concern, but maybe that should just be "vulnerable" directly.

People who pick a less-used browser or read a less-used language are more identifiable — need fewer other fingerprinting bits to identify them — than people who use more-popular languages and browsers. Similarly, if you turn on an experimental feature in your browser, that makes you much more identifiable. I don't think we can do anything about that, and our principles shouldn't say we have to do impossible things.

There's also an aspect of making privacy invasions less cost-effective. If actors can't amortize across as many victims, fewer will be able to afford to develop privacy attacks.

> * Editorial: change “fundamental aspects of the web" to “fundamental architectural features of the web such as interoperability”

This'll need to match the above decision about "architecture". I might use "linking" as the example if we keep the architectural limitation?

> * Discuss: could we avoid using the “trade-off” terminology?

Indeed, we should discuss this. I think it's very important to acknowledge that we're making trade-offs. :)

> * Re: “and users have come to expect this behavior,” add "in such contexts" because user's would not always have that expectation.

+1 that it's not in all contexts, but I think "such" isn't quite right, because I haven't introduced the contexts in this paragraph. Do you mean "in web browsers" or "on some websites" or something else?

> * Re: "When a violation occurs at their divisions between different browser profiles or at the point a user clears their cookies and site storage .." consider adding other signals such as use of extensions or add-ons to blocking tracking and private or incognito mode.

How about "When a violation occurs at their other divisions, for example between different browser profiles or at the point a user clears their cookies and site storage"? We should add the extensions and add-ons to the definition of "partition", rather than at each place that uses it. 

> * Re: "The web platform offers many ways for a website to recognize that a user is using the same identity over time .." - does this include ways for a website to recognise that a user has changed identity but is the same user?

I don't think so.

> * Discuss: "A privacy harm occurs if the user expresses that they intend ..". I understand the point of trying to tie the privacy harm to something that has been positively communicated by the user, but a privacy harm can occur if what happens does not match a user's expectation (whether or not they actively expressed their intent). Perhaps it would be helpful if I had a better understanding of what is meant by "if the user expresses that they intend ..".

This makes sense. I intuitively feel that declaring something a "privacy harm" means someone is to blame for the harm, and it's hard to blame someone if the user didn't express anything to the entity that caused the harm. But there might be something about user expectations that sites should know to expect.

> * Discuss: user agent responsibility - (and putting that aside for the moment, I will make the observation that at least some browsers (if not all major browsers) offer users' at least some level of protection against same-site recognition.

I think they're not automatically protecting against the intra-site context boundaries this paragraph talks about, but we can discuss the details and how to express them.

> * Editorial: change "personally-identifying information" to "personal data" or "personal information"

Here I do mean information that identifies a person, like an email address or ID number. Entering just a birth date would be personal information but not enough to join identities on its own. Is there a better term to avoid the "Everyone's wrong when they write PII" reflex?

index.html
```diff
@@ -902,32 +902,62 @@
...
+* This principle uses "quickly" because it currently appears impossible to prevent some forms of
+  fingerprinting that take a long time or many visits within each [=partition=].
+* This principle allows violations that only affect a small fraction of people who use the web
+  because people need to be allowed to configure their systems in unique ways, for example by using
+  a browser with a very small number of users. As long as a tracker can't track a significant number
+  of people, it's likely to be unviable to maintain the tracker. However, this doesn't excuse making
+  small groups of people trackable when those people didn't choose to be in the group.
+* This principle also allows violations when preventing those violations would break fundamental
+  aspects of the web. This document delegates to other documents, for example the
+  [[[PRIVACY-THREAT]]], to discuss which tradeoffs to make on this point.
```
jyasskin, line 934: Mention that in a lot of cases, a better design can achieve both goals.

index.html
```diff
@@ -902,32 +902,62 @@
...
+example between different browser profiles or at the point a user clears their cookies and site
+storage, that leads to <a href="#hl-recognition-same-site"></a>.
+
+### Same-site recognition {#hl-recognition-same-site}
+
+The web platform offers many ways for a website to recognize that a user is using the same
+[=identity=] over time, including [[RFC6265|cookies]], {{WindowLocalStorage/localStorage}},
+{{WindowOrWorkerGlobalScope/indexedDB}}, {{CacheStorage}}, and other forms of storage. This allows
+sites to save the user's preferences, shopping carts, etc., and users have come to expect this
+behavior.
```
jyasskin, line 948:
```suggestion
behavior in some contexts.
```

index.html
```diff
@@ -902,32 +902,62 @@
...
+
+A privacy harm occurs if the user expresses that they intend to use a different [=identity=] on a
```
jyasskin, line 950:
```suggestion
A privacy harm occurs if the user reasonably expects that they'll be using a different [=identity=] on a
```

index.html
```diff
@@ -902,32 +902,64 @@
...
+* This principle allows violations that only affect a small fraction of people who use the web
+  because people need to be allowed to configure their systems in unique ways, for example by using
+  a browser with a very small number of users. As long as a tracker can't track a significant number
+  of people, it's likely to be unviable to maintain the tracker. However, this doesn't excuse making
+  small groups of people trackable when those people didn't choose to be in the group.
+* This principle also allows violations when preventing those violations would break fundamental
+  aspects of the web. In many cases it's possible to change the design in a way that avoids the
+  violation without breaking valid use cases, but for cases where that's not possible this document
+  delegates to other documents, for example the [[[PRIVACY-THREAT]]], to discuss what detailed
+  tradeoffs to make.
```
jyasskin, line 936: How's this @sandandsnow?

jyasskin (commented):
I think the only remaining comment is about whether to talk about sites collecting a piece of "personally-identifying" or "personal" information in order to join identities between sites.

index.html
sandandsnow: @jyasskin, thank you for this revision. I think this works, though I still would prefer if we could find a way to avoid saying "the principle allows violations ..." in the document and instead say something along the lines of "in very limited circumstances, such as xyz, it may be impossible to guarantee that the effect of this principle applies to all users". (That wording probably is not right either, but I would prefer to move away from "allowing violations".

sandandsnow:
Regarding 3.1.2 Unwanted cross-site recognition, it seems we need a privacy statement like that in 3.1.1 Same-site recognition. Perhaps something along these lines "A privacy harm occurs if the user reasonably expects that a site cannot or will not discover that they came to that site from another known site or sites, but the site discovers and uses that information anyway." Also, a Best Practice.

jyasskin:
I think this now catches everything we discussed this week.

index.html
```diff
@@ -902,32 +902,64 @@
...
+As described in [[[#identity]]], [=cross-context recognition=] can sometimes be [=appropriate=], but
+users need to be able to control when websites do it as much as possible.
 
-A privacy harm only occurs if the user wants to break the association between
-two visits, but the site can still determine with high probability that the two
-visits came from the same user.
+<div class="practice">
 
-A user might expect that their two visits won't be associated if they:
+<p><span class="practicelab" id="principle-prevent-cross-partition-recognition">[=User agents=]
```
hober, line 914: (Sorry for the ReSpec ignorance, but) can't this be simplified to `<p class="practicelab ...>`? Why the nested span?

index.html
```diff
@@ -902,32 +902,64 @@
...
+  small groups of people trackable when those people didn't choose to be in the group.
+* This principle is also limited in cases where preventing recognition would break fundamental
+  aspects of the web. In many cases it's possible to change the design in a way that avoids the
+  violation without breaking valid use cases, but for cases where that's not possible this document
+  delegates to other documents, for example the [[[PRIVACY-THREAT]]], to discuss what detailed
+  tradeoffs to make.
+
+[=Partitions=] are separated in two ways that lead to distinct kinds of user-visible recognition.
+When their divisions between different sites are violated, that leads to <a
+href="#hl-recognition-cross-site"></a>. When a violation occurs at their other divisions, for
```
hober, line 940: This link is empty.

hober (approved):
Looks good! Here are some nits.

index.html
```diff
@@ -902,32 +902,64 @@
...
+</div>
 
-This recognition is generally accomplished by either "supercookies" or <a
-data-cite="fingerprinting-guidance#dfn-browser-fingerprinting">browser
+* This principle uses "probably" because websites can do harm even if they can't be completely
+  certain that visits come from the same person.
+* This principle uses "quickly" because it currently appears impossible to prevent some forms of
+  fingerprinting that take a long time or many visits within each [=partition=].
+* This principle is limited in cases that only affect a small fraction of people who use the web
+  because people need to be allowed to configure their systems in unique ways, for example by using
```
wseltzer, line 928: instead of "people need to be allowed to" 
```suggestion
  because people may configure their systems in unique ways, for example by using
```

torgo (approved):
looks good with applied changes from today's session

Issue #26, opened 2021-07-30
Definitions 2.1 to 2.4 (part)


sandandsnow:
**Person:** A suggestion to replace "person" with "individual" because, at least in some jurisdictions, the term "person" or "persons" could include corporations or bodies politic
**Personal Data:** Definitions of "personal data" or "personal information" are generally not limited to "identifiers". While the use of identifiers, especially persistent and globally-unique, are of particular privacy concern in Web standards, I do not think the definition should be this narrowly scoped. Separately, consider adding the scenario where a user can be singled out, even if not identified as X.
**Pseudonymous:** Review this definition. For instance, I could imagine a scenario where the identifiers used in the data  under the direct and exclusive control of the first party are not in fact pseudonymous (e.g., an email address that contains the user's name). Is the purpose to define what pseudonymous identifiers are, rather than pseudonymous data?
**Vulnerable person:** Is there a reason why the definition is focused on whether an individual is capable of giving meaningful consent (or perhaps I misread the definition)? Perhaps there are other ways that individuals are vulnerable in terms of privacy (e.g., power-imbalance, personal attributes which might cause them to be easily singled-out or identified as a person with x characteristics, etc.)
**Party:** For clarification, why does the definition of a "party" include a set of legal entities? Is this because of the first party, third party distinction typically used in W3C documents? (They might be related parties, but they are separate legal entities, which may have different privacy obligations or entitlements to use "personal data". )
**First Party:** It's helpful to provide some examples of where the user does not intend to interact with the first party. Would it also be helpful to add some examples of where it is generally agreed that the user does intend to interact?
**Service Provider or Data Processor:** For clarification, why does the data processor processing data on behalf of the first party need to be treated as the first party? (also relevant to the definition of **shares**.
**Purpose:** If we are prescribing characteristics to "purpose" such as "specific", there are other characteristics that should be added such as "reasonable, necessary, fair".

torgo:
Reviewed and discussed in [today's call]().

darobin:
For **pseudonymous**: make it clearer that embedded data (as per @sandandsnow's example of a name in an email address) that could lead to reidentification should be covered by "technical means" and also make it clearer that we don't need there to be a third party.

darobin:
About **party**: make it stricter, and remove _set_ of legal entities.

**Vulnerable**: make it a lot less about consent and more about autonomy and asymmetries of power.

darobin:
[...]

PR #43, opened 2021-09-01
Make a "party" a single legal entity, not a set.

jyasskin:
This diverges from the https://www.w3.org/TR/tracking-dnt/ definition, which
does allow sets of legal entities to be a single party. These principles are
instead moving toward an approach where multiple parties can be
considered first party as long as they meet conditions like the ones deleted in
this change, which should satisfy the reasons DNT allowed multiple entities to
group into a single party.

This would make #32 unnecessary. Contributes to #26.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/43.html#parties" title="Last updated on Sep 1, 2021, 6:44 PM UTC (23a4e1e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/43/4d3d266...jyasskin:23a4e1e.html" title="Last updated on Sep 1, 2021, 6:44 PM UTC (23a4e1e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -346,7 +346,7 @@ intellectual or psychological impairment, or refugees.
 
 ## The Parties {#parties}
 
-A <dfn>party</dfn> is a [=person=] or a legal entity.
+A <dfn>party</dfn> is a [=person=] or other legal entity.
 
 The <dfn data-lt="first parties">first party</dfn> is a [=party=] with which the [=user=] intends to
 interact. Merely hovering over, muting, pausing, or closing a given piece of content does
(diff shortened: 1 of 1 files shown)


jyasskin:
@sandandsnow Done. :)

dmarti:
The way that web site user agreements are written today, often some web users who have an "account on example.com" are party to an agreement with Example.com Inc. of the USA, and others have an agreement with Example.com Ltd. in another country. This is a fairly common tax avoidance strategy. Requiring an actual single entity might be positive or negative overall, but would increase the scope of this document.

dmarti (changes requested):
This doesn't seem realistic considering how web site user agreements for large sites are often structured (a US user will be party to an agreement with a US legal entity, Europeans party to an agreement with an entity in Europe, etc.)

PR #55, opened 2021-10-06
Removing fiduciary

### LINKED ISSUE #36: §2.5: use a different term than "fiduciary" or drop it
"Fiduciary" has very particular legal meanings and implications which vary from jurisdiction to jurisdiction. Perhaps there's an alternative word we could use that isn't loaded in the same way? Or maybe the word can simply be dropped. We could talk about the duties of user agents without saying they're specifically fiduciary duties.

darobin:
This fixes #36.

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/pull/55.html#user-agents" title="Last updated on Oct 13, 2021, 4:26 PM UTC (b40c135)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/55/6eed4d4...b40c135.html" title="Last updated on Oct 13, 2021, 4:26 PM UTC (b40c135)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -207,6 +207,11 @@
           href: 'https://papers.ssrn.com/sol3/papers.cfm?abstract_id=2655719',
           authors: ['Neil Richards', 'Woodrow Hartzog'],
         },
+        'TWITTER-DEVELOPER-POLICY': {
+          title: 'Developer Policy - Twitter Developers',
+          href: 'https://developer.twitter.com/en/developer-terms/policy',
+          publisher: 'Twitter'
+        },
         'TRACKING-PREVENTION-POLICY': {
           title: 'Tracking Prevention Policy',
           href: 'https://webkit.org/tracking-prevention-policy/',
@@ -277,8 +282,8 @@ understanding is that the law is a floor, not a ceiling.
 
 # Definitions {#definitions}
 
-This section provides a number of elementary building blocks from which to establish a
-shared understanding of privacy. Some of the definitions below build atop the work in
+This section provides a number of building blocks to create a shared understanding of
+privacy. Some of the definitions below build on top of the work in
 <em>Tracking Preference Expression (DNT)</em> [[tracking-dnt]].
 
 ## People &amp; Data {#people-data}
@@ -513,6 +518,9 @@ monitoring of their behaviour. Its <dfn>user agent duties</dfn> include
     transparency that is often included in older privacy regimes. Unlike transparency, honesty
     can't hide relevant information in complex out-of-band legal notices and it can't rely on
     very short summaries provided in a consent dialog.
+    If the user has provided [=consent=] to [=processing=] of their [=personal data=],
+    the [=user agent=] should inform the user of ongoing [=processing=], with a
+    level of obviousness that is proportional to the reasonably foreseeable impact of the processing.
   </dd>
   <dt><dfn>Duty of Loyalty</dfn></dt>
   <dd>
@@ -559,9 +567,9 @@ store data about that user.
 
 <div class="practice">
 
-<p><span class="practicelab" id="principle-identity-per-context">[=User agents=] should support
+<span class="practicelab" id="principle-identity-per-context">[=User agents=] should support
 their [=users=]' [=autonomy=] by helping them present their intended [=identity=] to each
-[=context=] that they visit.</span></p>
+[=context=] that they visit.</span>
 
 </div>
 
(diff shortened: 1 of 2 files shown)


darobin:
This is good to merge if @jyasskin likes it.

index.html
```diff
@@ -475,64 +475,67 @@
...
+[=user agent=] is <em>not</em> a [=context=] in that it is expected to align fully with its
+[=user=] and operate exclusively in that [=person=]'s interest. It is <em>not</em> the
+[=first party=]. The [=user agent=] serves the [=user=] as a <dfn>trustworthy agent</dfn>:
+it always puts the [=user=]'s interest first. In some occasions, this can mean protecting
+the [=user=] from themselves by preventing them from carrying out a dangerous decision,
+or at the very least by slowing down the user in their decision. For example, the
+[=user agent=] will make it difficult for the [=user=] to connect to a site if it can't verify
+that the site is authentic. It will check that the [=user=] really intends to expose a
+sensitive device to a page. It will prevent the [=user=] from consenting to the permanent
+monitoring of their behaviour. Its <dfn>user agent duties</dfn> include
```
jyasskin, line 487: Related to #62, the UA won't know about all of the user's consents or all the monitoring around them, but we could probably be more precise to get at something the UA can do here. No need to fix this in this PR.

index.html
```diff
@@ -475,64 +475,67 @@
...
   <dt><dfn>Duty of Honesty</dfn></dt>
   <dd>
-    Honesty requires that the [=user agent=] make sure that the [=user=] is proactively
-    provided with information that is relevant to them and that will enhance the [=user=]'s
-    autonomy, to the extent possible in a manner that they will comprehend and at the right
-    moment, which is almost never when the [=user=] is trying to do something else such as
+    Honesty requires that the [=user agent=] make sure that the [=user=] is actively
+    given information that is relevant to them and that will increase the [=user=]'s
+    autonomy, as much as possible in a manner that they will comprehend and at the right
+    moment. This is almost never when the [=user=] is trying to do something else such as
```
jyasskin, line 511:
```suggestion
    Honesty requires that the [=user agent=] try to give the [=user=]
    information that is relevant to them and that will increase the [=user=]'s
    autonomy, as long as they can understand it and there's an appropriate
    time. This is almost never when the [=user=] is trying to do something else such as
```

index.html
```diff
@@ -475,64 +475,67 @@
...
-    including in preference to the [=user agent=]'s implementer. When a [=user agent=] carries
-    out [=processing=] that is not directly in the [=user=]'s interest but rather benefits
-    another entity such as its implementer, including by piggybacking on [=processing=] that
-    may be in the user's interest, that behaviour is known as <dfn>self-dealing</dfn>.
-    [=Self-dealing=] is always [=inappropriate=]. Loyalty is the avoidance of [=self-dealing=].
+    Because the [=user agent=] is a [=trustworthy agent=], it is held to be loyal to the
+    [=user=] in all situations, including in preference to the [=user agent=]'s implementer.
+    When a [=user agent=] carries out [=processing=] that is not directly in the [=user=]'s
+    interest but rather benefits another entity such as the implementer that behaviour is
+    known as <dfn>self-dealing</dfn>. Behaviour is [=self-dealing=] even if it is done at the
```
jyasskin, line 523: I'll quibble with "directly", but we can figure that out in another PR. Here's an attempt to simplify the meaning you already have:

```suggestion
    When a [=user agent=] carries out [=processing=] that is not directly in the [=user=]'s
    interest but instead benefits another entity (such as the user agent's implementer) that behaviour is
    known as <dfn>self-dealing</dfn>. Behaviour can be [=self-dealing=] even if it is done at the
```

index.html
```diff
@@ -475,64 +475,67 @@
...
 </dl>
 
 These duties ensure the [=user agent=] will <em>care</em> for the user. It is important to
-note that there is a subtle difference between care and <dfn>data paternalism</dfn> which
-is that the latter purports to help in part by removing agency ("<em>don't worry about it,
+note that there is a subtle difference between care and <dfn>data paternalism</dfn>.
+[=Data paternalism=] claims to help in part by removing agency ("<em>don't worry about it,
 so long as your data is with us it's safe, you don't need to know what we do with it, it's
-all good because we're good people</em>") whereas care aims to support people by enhancing
+all good because we're good people</em>") whereas [=care=] aims to support people by enhancing
```
jyasskin, line 533: There's no `<dfn>` for "care".
```suggestion
all good because we're good people</em>") whereas care aims to support people by enhancing
```

index.html
```diff
@@ -475,64 +475,67 @@
...
+    [=inappropriate=]. Loyalty is the avoidance of [=self-dealing=].
   </dd>
 </dl>
 
 These duties ensure the [=user agent=] will <em>care</em> for the user. It is important to
-note that there is a subtle difference between care and <dfn>data paternalism</dfn> which
-is that the latter purports to help in part by removing agency ("<em>don't worry about it,
+note that there is a subtle difference between care and <dfn>data paternalism</dfn>.
+[=Data paternalism=] claims to help in part by removing agency ("<em>don't worry about it,
 so long as your data is with us it's safe, you don't need to know what we do with it, it's
```
jyasskin, line 532: This isn't about what you're fixing in this PR, so no need to change it in this PR, but this paraphrase for "data paternalism" seems more about conning people into giving up their data. Paternalism would be more like "you can't enter your address into this form because we think its owner will mis-use it." or "you can't connect to this site because its TLS is broken and we don't trust you to make a good risk tradeoff.", right?

PR #78, opened 2021-11-04
Allow UA behavior that only indirectly benefits the user.

jyasskin:
In particular, crash reporting, use counters, and similar telemetry only
benefit users in aggregate, which doesn't satisfy the original wording here.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/78.html#dfn-duty-of-loyalty" title="Last updated on Nov 5, 2021, 4:42 AM UTC (74d4ed4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/78/75e1d19...jyasskin:74d4ed4.html" title="Last updated on Nov 5, 2021, 4:42 AM UTC (74d4ed4)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -529,7 +529,7 @@ monitoring of their behaviour. Its <dfn>user agent duties</dfn> include
   <dd>
     Because the [=user agent=] is a [=trustworthy agent=], it is held to be loyal to the
     [=person=] using it in all situations, including in preference to the [=user agent=]'s implementer.
-    When a [=user agent=] carries out [=processing=] that is not directly in the [=person=]'s
+    When a [=user agent=] carries out [=processing=] that is not in the [=person=]'s
     interest but instead benefits another entity (such as the user agent's implementer) that behaviour is
     known as <dfn>self-dealing</dfn>. Behaviour can be [=self-dealing=] even if it is done at the
     same time as [=processing=] that is in the [=person=]'s interest. [=Self-dealing=] is always


wseltzer:
In this context (with reference to duty of loyalty), I agree with the change. It's worth thinking elsewhere about giving visibility to the difference between direct and indirect benefits.

chrisn:
I also think it's worthwhile being more explicit than simply removing the word "directly". If some features are there to benefit website developers, and hence only indirectly benefit end users, then this document should call that out explicitly.

dmarti:
I also agree with this change. For a complicated transfer of data like a crash report, the user agent implementer should be considering what information the user would choose to share if they took the time to understand.

(The user should be able to opt in or out of this kind of behavior if they are one of the few people who want to actually micro-manage it, but defaults can be set based on the results of user research.)

darobin:
I think I agree with the intent of the change but I object to the actual change.

The way I see it, UA behaviour should either _directly_ benefit the user or it should abide by specific precepts of collective governance (ideally even be in a list of well-defined exceptions the governance model for which is the group that owns this document) that this document should list clearly. So for instance we could have (this is just quick drafting for flavour, I'm sure a lot is wrong with it):

* Crash reporting: the data can be quite invasive, but must only be collected for the purpose of fixing bugs, retention must be limited to 3 months, and only people directly involved in fixing bugs must be granted access to it.
* Use telemetry: the data can only be collected in privacy-preserving ways that adhere to certain standards (eg. Prio) and must be made publicly usable.
* Safety: the data can only be collected in privacy-preserving ways yada yada and must be under collective governance amongst browsers since safety is a public good.

The concern I have is that the second you open the door is _indirect_ benefits, pretty much everything could theoretically be good for the user. We all know people who won't hesitate to claim that extensively tracking users is good, actually, because that's the value exchange that "finances journalism." Without going into caricatural positions, collecting data to improve an unrelated service could be argued to be good for the user (the service is better for those who use it) and bad for the user (it's leveraging the browser to bolster dominance in a different space, meaning better, more meritorious alternatives that users might have preferred do not get to exist). That's an endless rabbit hole.

So I would say yes to introducing indirection, but we should do it in such a way that we have consensus over what indirect goods are actually good.

jyasskin:
I'm happy to refine the way we allow "good" indirect benefits.

I think this document should describe the principles for how we balance benefits to different parties, rather than itself trying to include a complete list of times indirect benefits are allowed.

In Robin's examples, I see a couple possible principles (with the same caveat that I haven't thought these through deeply):
* Data collected from a person should benefit that person (directly or indirectly) and shouldn't also unfairly benefit someone else. (For example, a 1 [util](https://en.wiktionary.org/wiki/util#English) benefit to a person would generally not justify a 100 util benefit to the browser maker, but could justify a 1 util benefit to the browser maker.)
* Data about a person that doesn't directly benefit that person should almost always be permanently de-identified.
   * When it can't be (like with crash reports), it needs to be tightly controlled so it doesn't wind up harming the person or unfairly benefitting someone else.
* Data that only indirectly benefits the person it was collected from should be made publicly usable.
* Data use with indirect benefits should be governed by a multistakeholder process.

These principles seem to cover the other example I came up with, the [Chrome User Experience Report](https://developers.google.com/web/tools/chrome-user-experience-report), except that we don't have a multistakeholder process. There are probably even more examples that'll illustrate holes.

PR #80, opened 2021-11-05
Say that careful cross-context data sharing can be appropriate.

jyasskin:
My thought here is that offline data sharing between contexts, even contexts with very different norms, happens all the time, and the people who do it are just careful to avoid breaking the norms of the context they got the data from. Online, the norms of two different contexts could easily be similar enough to allow a lot of cross-context use without breaking any user expectations or preferences.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/80.html#contexts-and-privacy" title="Last updated on Nov 24, 2021, 5:31 PM UTC (b464961)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/80/4056f53...jyasskin:b464961.html" title="Last updated on Nov 24, 2021, 5:31 PM UTC (b464961)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -473,16 +473,19 @@ We define <dfn>privacy</dfn> as a right to [=appropriate=] [=data processing=]. 
 
 Note that a [=first party=] can be comprised of multiple [=contexts=] if it is large
 enough that [=people=] would interact with it for more than one [=purpose=]. [=Sharing=]
-[=personal data=] across [=contexts=] is, in the overwhelming majority of cases,
-[=inappropriate=].
+[=personal data=] across [=contexts=] is only [=appropriate=] if the subsequent
+use of that data follows the [=norms=] of the context in which the data was collected.
 
 <div class="example">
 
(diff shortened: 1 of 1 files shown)


dmarti:
I can see how this could work as long as the data sharing follows the norms of both the source and destination context. A user might have data collected in a context with high personalization norms but then move to a context with low personalization norms. Maybe instead of 

> Sharing personal data across contexts is only appropriate if the subsequent use of that data follows the norms of the context in which the data was collected.

it could be something like 

> Sharing personal data across contexts is only appropriate if the subsequent use of that data follows the norms of the context in which the data is used and the norms of the context in which the data was collected.

jyasskin:
Yeah, I wasn't sure whether to require just the norms of the collecting context or both contexts (or even _all_ contexts if data is shared across multiple hops). I think systems in practice would have to follow the union of at least the source and final-destination norms in order to have a chance of keeping the restrictions straight, and I'm fine with requiring that.

PR #94, opened 2022-01-07
Reorganize the document to flow better

### LINKED ISSUE #89: Organize the document coherently
Since our document started as a concatenation of two independent documents, at some point we'll need to refactor the whole thing into a coherent whole. I'm starting to get a picture of how we might do that, which I'll outline here.

1. [Introduction](https://w3ctag.github.io/privacy-principles/#intro)
1. Interesting Definitions
   1. [People](https://w3ctag.github.io/privacy-principles/#people)
   1. [Data](https://w3ctag.github.io/privacy-principles/#data)
   1. [Parties](https://w3ctag.github.io/privacy-principles/#parties)
   1. [Acting on Data](https://w3ctag.github.io/privacy-principles/#acting-on-data)
   1. "purpose" and "means" definitions, from later sections.
1. [User Agents](https://w3ctag.github.io/privacy-principles/#user-agents)
1. [Identity](https://w3ctag.github.io/privacy-principles/#identity)
   1. [Cross-context recognition](https://w3ctag.github.io/privacy-principles/#hl-recognition-cross-context), possibly mixed together with the Identity section
1. Ethical Data Use
   1. [Personal Control and Autonomy](https://w3ctag.github.io/privacy-principles/#autonomy)
   1. [Contexts](https://w3ctag.github.io/privacy-principles/#contexts-and-privacy)
   1. [Opt-in, Consent, Opt-out, Global Controls](https://w3ctag.github.io/privacy-principles/#opt-in-out)
   1. #73
   1. [Sensitive information disclosure](https://w3ctag.github.io/privacy-principles/#hl-sensitive-information)
   1. [Unexpected profiling](https://w3ctag.github.io/privacy-principles/#hl-unexpected-profiling)
   1. [Collective issues](https://w3ctag.github.io/privacy-principles/#collective)
1. [Intrusive behavior](https://w3ctag.github.io/privacy-principles/#hl-intrusion)
1. [Powerful capabilities](https://w3ctag.github.io/privacy-principles/#hl-capabilities)
1. Boring Definitions

I think we could further reorganize the sections under "Ethical Data Use" by thinking about how the grounds for lawful processing, for example from the [GDPR](https://gdpr-info.eu/art-6-gdpr/) map to the principles we'd like organizations to use when deciding how to process data. For example, data use is ok if it's clear to the user that it's necessary to do what they asked; freely-given consent makes data use ok; data use to protect user safety is generally ok; use that's in-line with contextual expectations is generally ok; etc.

jyasskin:
I tried to incorporate Robin's goal for a coherent narrative instead of just a specification, and my goal of including non-data privacy. I left lots of issues where there's still more work to do to clean up the text, but hopefully the direction is clear and matches y'all's intentions.

Fixes #89.

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/pull/94.html" title="Last updated on Jan 12, 2022, 7:17 PM UTC (9479374)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/94/9fd56a9...9479374.html" title="Last updated on Jan 12, 2022, 7:17 PM UTC (9479374)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1152,8 +1152,6 @@ in a preferences store (means)</em>.
 
 </section>
 
-<section class="appendix" id="bp-summary"></section>
-
 <section class="appendix">
 
 # High-Level Threats {#threats}
@@ -1234,6 +1232,9 @@ These threats combine into the particular concrete threats we want web
 specifications to defend against, described in the sections that follow.
 
 </section>
+
+<section class="appendix" id="bp-summary"></section>
+
 <section class="appendix">
 
 # Acknowledgements {#acknowledgements}
--- a/index.html
+++ b/index.html
@@ -274,12 +274,23 @@ conclusion of the task force, the TAG intends to adopt this document as a Findin
 
 # Introduction {#intro}
 
-Privacy is an essential value of the Web ([[?ETHICAL-WEB]], [[?design-principles]]).
+<dfn>Privacy</dfn> is an essential value of the Web ([[?ETHICAL-WEB]],
+[[?design-principles]]).
+
+<aside class="issue">
+
+Improve this introduction to introduce the several forms of privacy covered in this
+document. The term privacy covers several related topics, both about having <a
+href="#data-privacy">reasonable control over other people's impressions of you</a>, and <a
+href="#intrusion">preventing them from bothering you</a>. Privacy can be <dfn
+data-lt="privacy violation">violated</dfn>.
+
+</aside>
 
 The goal of this document is to define principles that may prove useful in developing
 technology and policy that relate to privacy and [=personal data=].
 
-[=Personal data=] is covered by legal frameworks and this document recognises that existing data
+Privacy is covered by legal frameworks and this document recognises that existing data
 protection laws take precedence for legal matters. However, because the Web is global, we
 benefit from having shared concepts to guide its evolution as a system built for the people using it
 [[?RFC8890]]. A clear and well-defined view of privacy on the Web, informed by research,
@@ -290,21 +301,21 @@ understanding is that the law is a floor, not a ceiling.
 
 ## People {#people}
 
-A <dfn data-lt="individual">person</dfn> (also <dfn>user</dfn> or <dfn>data subject</dfn>) is any natural
+A <dfn data-lt="individual|people">person</dfn> (also <dfn>user</dfn> or <dfn>data subject</dfn>) is any natural
 person. Throughout this document, we primarily use [=person=] or [=people=] to refer to human beings, as a
 reminder of their humanity. When we use the term [=user=], it is to talk about the specific [=person=]
 who happens to be using a given system at that time.
 
-A <dfn data-lt="vulnerable">vulnerable person</dfn>, in the [=context=] of a specific kind of
-[=data processing=], is a [=person=] who may be unable
-to exercise sufficient self-determination. Amongst other things, they should be treated with greater default
-privacy protections and may be considered unable to [=consent=] to the [=processing=].
+A <dfn data-lt="vulnerable">vulnerable person</dfn> is a [=person=] who may be unable to
+exercise sufficient self-determination in a [=context=]. Amongst other things, they should
+be treated with greater default privacy protections and may be considered unable to
+[=consent=] to various interactions with a system.
 People can be vulnerable for different reasons, for example because they are children,
 are employees with respect to their employers, are facing a steep asymmetry of power,
 are people in some situations of intellectual or psychological impairment, are
 refugees, etc.
 
-## User Agents {#user-agents}
+## Their Software Agents {#user-agents}
 
 The <dfn>user agent</dfn> acts as an intermediary between a [=person=] (its [=user=]) and the web. The
 [=user agent=] is <em>not</em> a [=context=] because it is expected to align fully with the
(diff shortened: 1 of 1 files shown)


darobin:
Thanks a lot for doing this!

I think, reading this laid out, that my primary reaction is that I think I am now convinced that we should restrict the scope to _information_ privacy. As such I think that we should drop Section 6 "How We Experience the World". My reasons for this are:

* This strikes me as better addressed in the context of "powerful capabilities," a topic that I think it worthy of its own considerations (HTTPS, visit frequency/recency, etc.).
* I can think of three broad views on powerful capabilities and I don't believe that we are in a position to drive consensus between these (legitimate, user-centric for different reasons) views, namely:
  1. Web sites should not have powerful capabilities because they are too dangerous and it is too hard to make the Web safe with them.
  2. Web sites should have powerful capabilities because they need to compete with apps, and we'll figure out how to handle it.
  3. Powerful capabilities are desirable, but we should only increase Web sites' power if and when we improve the Web's trust model.
* From a "serving our constituents" perspective, I see an immediate need for a principled take on information privacy (it's needed for PATCG, WebAdv, Privacy CG, the AC…) but I see no immediate need to tackle powerful capabilities. (This isn't to say that we shouldn't! Just that it doesn't strike me as being driven by a burning priority.)
* From a "building bridges to policy" perspective, there's a similar view. There _are_ some early regulatory proposals about dark patterns, but they generally don't cover the space that is considered here (and they probably shouldn't).
* We don't have content for this section at this time.

Again, I am emphatically _not_ saying that this isn't something that the TAG should do, I just don't think that there's a real opportunity in treating these two topics together, with the same group of people. In a sense, this "intrusive" set is quite similar to spatial privacy (your neighbour mowing their lawn at 7am Sunday morning, someone barging into the loo as you're using it, manspreading…). It's a cousin problem, but I think we might be better off not including the entire family.

If we remove that section, I assume that some of the rest of the doc would get rejiggled, though not necessarily that much? I think that much of the singled-out information privacy could move to the intro, which might help give it more narrative feel.

Smaller notes:
* I should note as well that information privacy isn't about "how the world sees me" but much more about "what leverage does a given other entity have over me."
* I think there's a fair chunk of the definitions that can be moved to the appendix. But that can be done later.

sandandsnow:
@jyasskin, thank you. What a lot of work. I think the revised structure is a good approach. I might put threats before the best practices, but I am okay with it staying in the same order. Are we still going to have a section of principles (basically the overarching principle(s) for the best practices)? Are we going to expand the threats section to provide some detail of specific threats in the Web technologies context?

Some comments on the content (from this read through):
* Re: section 1: “The goal of this document is to define principles that may prove useful in developing technology and policy that relate to privacy and personal data”, it would be helpful if we could elevate the goal to “define privacy principles for Web technologies” (or similar)
* Re: section 2.2: Perhaps we could leave out the point about data paternalism as the explanation of user agent duties make it clear that is not what is envisaged. Also, is “duty of discretion” the right term? You might also want to note at the end that “fiduciary” also has legal meaning.
* Re: section 3: One aspect to keep in mind is that typically in online contexts other parties may be present that are unobservable to the person. At an in-person cocktail party, a person can see who is present and may be monitoring their attributes and actions. Of course, today a cocktail party attendee may not know that the host has a so-called “smart assistant” in listening/observing mode.
* Re: section 4: Would we say burying privacy implications in pages of legalese is a dark pattern or another form of bad behaviour?
* Re: section 4.1: We might want to include a statement to the effect that some controls are in practice illusory because the person or user effectively has no choice, accept the terms or don’t use the service.
* Re: section 4.2: I would argue that privacy labour has been increased by a consent-based model for allowing data processing, i.e. where only consent is needed.
* (still reviewing content)

jyasskin:
A couple comments ahead of the meeting tomorrow:

* Re @darobin, I don't think the "How We Experience the World" aspect of privacy is primarily about powerful capabilities, and I actually removed the section that had been about powerful capabilities. The examples left in the "experience" section are all ones with cross-engine consensus, even [push](https://developer.apple.com/notifications/safari-push-notifications/) [messages](https://blog.pushpad.xyz/2021/10/apple-is-working-on-web-push-notifications-for-safari/).
* Re @sandandsnow:
   * This was less work than the size of the diff would indicate, since I could mostly cut&paste things into place. :)
   * I like your section-by-section comments, but I'll plan to do them after the big rearrangement is in, to let us focus on them individually.
   * I've flipped the threats and the best practices.
   * Yes, I believe we do still plan to have a longer list of principles. I think what are currently labeled "Best Practices" are those principles, but we need to extract more of them from the text. We'll rename that section once Respec supports it.
   * I'm not sure if the "high-level threats" section will survive to the published document, but I think we can consider that independently of this change.

torgo:
Maybe we should defer a decision on the "how we experience the world" section? I'm broadly supportive of it but I also feel we have more work to do on it.

darobin:
I think what's throwing me with having both information and spatial privacy in there is that it sort of forces sections 2-4 to be factored out as providing common grounding, which in turn makes the narrative a bit harder to land. (It feels like a lot of backstory, to stick to the fiction analogy.) Also, I don't think that the "world" in both section titles is the best angle on either (or at least on 5) and so I don't think it works as a way to create consistency between the two.

However, if @jyasskin thinks he can make it work my instinct would be to trust him and try. I like a fair bit of the other reorg and I think the bumps can be smoothed out. With that in mind I propose the following plan:

1. We use the call to go through @sandandsnow's feedback (and other 
2. We merge this with the changes from the previous step.
3. I take a stab at narrativising (sue me) further in a way that I _think_ could work with both what I have in mind and @jyasskin's vision. (I think that the trick might be to fold sections 1-4 into a single section that can support the other two but also does work on its own.)

Issue #97, opened 2022-01-12
Write example of about context violation


darobin:
This comment from @sandandsnow on #94 needs to be written up as a good example:

```
Re: section 3: One aspect to keep in mind is that typically in online contexts other parties may be present that are unobservable to the person. At an in-person cocktail party, a person can see who is present and may be monitoring their attributes and actions. Of course, today a cocktail party attendee may not know that the host has a so-called “smart assistant” in listening/observing mode.
```

Issue #35, opened 2021-08-17
Re: Opt-in, Consent, Opt-out, Global Controls


sandandsnow:
* First sentence: Change “control” to “express a preference for what processing is done to their data”. 
* Second sentence: “uses” would seem more relevant in this context than “purposes”, or perhaps “uses and/or purposes”.
* Informative website regarding dark patterns: https://darkpatternstipline.org/harms
* For discussion: the default privacy tiers. For example, it does not seem appropriate to require people who are vulnerable to opt-out. Consider also adding text about the privacy implications of choosing to opt-out, especially when the majority of users does not.
* Consider adding text about people rarely changing the default and the consequences that has for their privacy.
* Permissions are relevant to consent. Take a look at Nick Doty’s post here: https://www.w3.org/blog/2019/07/adding-another-permission/. It could be a helpful starting point for recommendations. Would it make sense to add another section under 3 that covers permissions?

jyasskin:
In https://github.com/w3ctag/privacy-principles/pull/94#issuecomment-1010535916, @sandandsnow also mentions

> We might want to include a statement to the effect that some controls are in practice illusory because the person or user effectively has no choice, accept the terms or don’t use the service.

PR #101, opened 2022-01-12
Fix some of sandandsnow's comments from #94.

jyasskin:
> Re: section 1: “The goal of this document is to define principles that may prove useful in developing technology and policy that relate to privacy and personal data”, it would be helpful if we could elevate the goal to “define privacy principles for Web technologies” (or similar)

Done.

> Re: section 2.2: Perhaps we could leave out the point about data paternalism as the explanation of user agent duties make it clear that is not what is envisaged.

Done, and I removed the reference from the autonomy section.

> Also, is “duty of discretion” the right term?

This comes from [Taking Trust Seriously in Privacy Law](https://papers.ssrn.com/sol3/papers.cfm?abstract_id=2655719), which is cited above the list of duties. I've tried to adjust the text to match that definition of discretion better, but @darobin please check that I haven't subverted your intent.

> You might also want to note at the end that “fiduciary” also has legal meaning.

Added. Please check that I said the right thing about the legal meaning.

> Re: section 3: One aspect to keep in mind is that typically in online contexts other parties may be present that are unobservable to the person. At an in-person cocktail party, a person can see who is present and may be monitoring their attributes and actions. Of course, today a cocktail party attendee may not know that the host has a so-called “smart assistant” in listening/observing mode.

Delegated to #97.

> Re: section 4: Would we say burying privacy implications in pages of legalese is a dark pattern or another form of bad behaviour?

We think it is in some cases, but there wasn't any text to add about this.

> Re: section 4.1: We might want to include a statement to the effect that some controls are in practice illusory because the person or user effectively has no choice, accept the terms or don’t use the service.

Delegated to #35.

> Re: section 4.2: I would argue that privacy labour has been increased by a consent-based model for allowing data processing, i.e. where only consent is needed.

Added.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/101.html" title="Last updated on Jan 19, 2022, 5:20 PM UTC (7568fe7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/101/36e9b27...jyasskin:7568fe7.html" title="Last updated on Jan 19, 2022, 5:20 PM UTC (7568fe7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -374,7 +374,7 @@ monitoring of their behaviour. Its <dfn>user agent duties</dfn> include
 
 These duties ensure the [=user agent=] will <em>care</em> for its [=user=]. In academic
 research, this relationship with a [=trustworthy agent=] is often described as "fiduciary"
-[[?FIDUCIARY-UA]]. Various legal systems have a similar meaning for "fiduciary duties".
+[[?FIDUCIARY-UA]]. Some jurisdictions may have a distinct legal meaning for "fiduciary".
 
 ## Guardians and Device Owners {#guardians-and-owners}
 
--- a/index.html
+++ b/index.html
@@ -605,7 +605,7 @@ form upon submission).
 Additionally, <dfn data-lt="privacy-labor">privacy labour</dfn> is the practice of having
 a [=person=] carry out the work of ensuring [=data processing=] of which they are the
 subject is [=appropriate=], instead of having the [=parties=] be responsible for that
-work as is more respectable. Data systems that are based on asking [=people=] for their
+work. Data systems that are based on asking [=people=] for their
 [=consent=] tend to increase [=privacy labour=].
 
 # How the World Sees Us {#data-privacy}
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -614,7 +605,8 @@
 Additionally, <dfn data-lt="privacy-labor">privacy labour</dfn> is the practice of having
 a [=person=] carry out the work of ensuring [=data processing=] of which they are the
 subject is [=appropriate=], instead of having the [=parties=] be responsible for that
-work as is more respectable.
+work as is more respectable. Data systems that are based on asking [=people=] for their
```
samuelweiler, line 608: what can we say instead of "more respectable"?

samuelweiler: opened #103

torgo, line 608:
```suggestion
work. Data systems that are based on asking [=people=] for their
```

torgo (approved):
lgtm

PR #105, opened 2022-01-21
Discuss tensions between principles.

### LINKED ISSUE #73: How do safety systems relate to privacy
We should consider things that use attack pathways for pro-user, defensive purposes, and how we should reason about it:

* Detecting tags/beacons in the surrounding space in order to report that a person may be tracked by someone else.
* Fraud prevention uses, for invalid traffic detection
* Relying on geolocation to check that a person's credit card purchases are valid

jyasskin:
Fixes #73.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/105.html#balancing" title="Last updated on Mar 30, 2022, 12:35 AM UTC (2447bd8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/105/e351f39...jyasskin:2447bd8.html" title="Last updated on Mar 30, 2022, 12:35 AM UTC (2447bd8)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1179,8 +1179,8 @@ improved privacy.
 ## Balancing Conflicting Privacy Rights {#balancing}
 
 Sometimes different people's privacy rights come into conflict. Often this happens when
-one user of a service is trying to invade another user's privacy: hiding all private
-information about the first user can prevent the service from protecting the second user.
+one user of a service is trying to invade another user's privacy: keeping a user's
+information private can limit the ways that a service can protect another user.
 
 <div class="practice">
 
--- a/index.html
+++ b/index.html
@@ -1183,15 +1183,15 @@ to offer better protection by default? Given the contested nature of the [^a/pin
 the absence of a forcing function to support collective enforcement, the scheme failed to deliver
 improved privacy.
 
-## Balancing Conflicting Privacy Principles {#balancing}
+## Balancing Different Privacy Principles {#balancing}
 
-Sometimes different privacy principles come into conflict. Often this happens when one
+Sometimes different privacy principles come into tension. Often this happens when one
 user of a service is trying to invade another user's privacy: keeping a user's information
 private can limit the ways that a service can protect them or another user.
 
 However, the choice of which privacy principles to prefer is complex and depends heavily
 on the details of each particular situation. <span class="note">Note that people's privacy
-can also conflict with other important non-privacy concerns. As discussed in the
+can also be in tension with other important non-privacy concerns. As discussed in the
 [[[Ethical-Web]]], it's important to balance these concerns too, but this document doesn't
 discuss how to do that.</span>
 
--- a/index.html
+++ b/index.html
@@ -708,7 +708,89 @@ research, this relationship with a [=trustworthy agent=] is often described as "
 Many of the [=principles=] described in the rest of this document extend the [=user agent=]'s duties and
 make them more precise.
 
+## Balancing Different Privacy Principles {#balancing}
 
+Sometimes different privacy principles come into tension. Often this happens when one
+user of a service is trying to invade another user's privacy: keeping a user's information
+private can limit the ways that a service can protect them or another user.
+
+However, the choice of which privacy principles to prefer is complex and depends heavily
+on the details of each particular situation. <span class="note">Note that people's privacy
+can also be in tension with other important non-privacy concerns. As discussed in the
+[[[Ethical-Web]]], it's important to balance these concerns too, but this document doesn't
+discuss how to do that.</span>
+
+<aside class="example" id="example-sockpuppets" title="Sockpuppets">
+
+A person might want to sign up many accounts ("sockpuppets") or disguise the affiliation
+of individual-owned accounts
+("[astroturfing](https://en.wikipedia.org/wiki/Astroturfing)") on a service in order to
+trick other people into thinking a belief has more support than it really has. This
+violates the other people's rights to be free from manipulation.
+
+On the other hand, identifying everyone with enough detail to detect these cases tends to
+violate their rights to be free from [=surveillance=] and [=correlation=].
+
+</aside>
+
+<aside class="example" id="example-children" title="Children's Services">
+
+Children using a service (or their guardians) may want to ensure their interaction with
+that service is only visible to other children. To accomplish this, the service would need
+to verify all of its users' ages.
+
+However, verifying ages is potentially an invasive process, requiring automated facial
+recognition, collection of live images, or strongly identifying the children (e.g. via
+[[NIST-800-63A]]).
+
+</aside>
+
+<aside class="example" id="example-account-security" title="Account Security">
+
+Accounts with a service need to be protected more strongly than just with a username and
+password. Since a compromised account reveals all the private information stored in the
+account, this is a privacy issue and not just a security issue. Services often store the
+historical locations and machine characteristics that have accessed an account, in order
+to make it harder to log in from an unusual place or type of machine. They might also
+store other personal data, like a phone number or address, in order to check that a
+suspicious login actually comes from the real account owner.
+
+</aside>
+
+It is attractive to say that if someone violates the norms of a service they're using,
+then they sacrifice a proportionate amount of their privacy protections, but
+
+1. Often the service can only prevent the norm violation by also collecting data from
+   innocent users. This extra collection is not always [=appropriate=], especially if it
+   allows pervasive monitoring ([[RFC7258]], [[RFC7687]]).
+1. If a service owner wants to collect some extra data, it can be tempting for them to
+   define norms and proportionality that allow them to do so.
+
+<div class="practice">
+
+<p><span class="practicelab" id="principle-limited-collection-for-safety">If a service
+collects extra data from its users in order to protect those or other users, it must take
+extra technical and legal measures to ensure that this data can't be then used for other
+purposes, like to grow the service.</span></p>
+
+This is a special case of the more general principle that data can't be used for more
+[=purposes=] than the data's subjects understood it was being collected for.
+
+</div>
+
+<aside class="example" id="example-technical-legal-measures" title="Technical and Legal Measures">
+
+A site might segregate the data it collects for safety reasons from its business data by:
+
+1. Writing in its privacy policy that these types of data are kept separate. (A legal
+   measure)
+1. Using multi-party computation to ensure that its business side can't learn the
+   sensitive safety data unless both its safety side and a trusted third party collude. (A technical measure)
+1. Hiring a trusted auditor to publicly check that the algorithms are sufficient to
+   provide the guarantees they claim to provide, and that there isn't, in fact, any
+   collusion. (Another technical measure)
+
+</aside>
 
 
 # Principles for Privacy on the Web
@@ -1182,91 +1264,6 @@ and bounce tracking for others? What prevents a browsers from opting everyone ou
 to offer better protection by default? Given the contested nature of the [^a/ping^] attribute and
 the absence of a forcing function to support collective enforcement, the scheme failed to deliver
 improved privacy.
-
-## Balancing Different Privacy Principles {#balancing}
-
-Sometimes different privacy principles come into tension. Often this happens when one
-user of a service is trying to invade another user's privacy: keeping a user's information
-private can limit the ways that a service can protect them or another user.
-
-However, the choice of which privacy principles to prefer is complex and depends heavily
-on the details of each particular situation. <span class="note">Note that people's privacy
-can also be in tension with other important non-privacy concerns. As discussed in the
-[[[Ethical-Web]]], it's important to balance these concerns too, but this document doesn't
-discuss how to do that.</span>
-
-<aside class="example" id="example-sockpuppets" title="Sockpuppets">
-
-A person might want to sign up many accounts ("sockpuppets") or disguise the affiliation
-of individual-owned accounts
-("[astroturfing](https://en.wikipedia.org/wiki/Astroturfing)") on a service in order to
-trick other people into thinking a belief has more support than it really has. This
-violates the other people's rights to be free from manipulation.
-
-On the other hand, identifying everyone with enough detail to detect these cases tends to
-violate their rights to be free from [=surveillance=] and [=correlation=].
-
-</aside>
-
-<aside class="example" id="example-children" title="Children's Services">
-
-Children using a service (or their guardians) may want to ensure their interaction with
-that service is only visible to other children. To accomplish this, the service would need
-to verify all of its users' ages.
-
-However, verifying ages is potentially an invasive process, requiring automated facial
-recognition, collection of live images, or strongly identifying the children (e.g. via
-[[NIST-800-63A]]).
-
-</aside>
-
-<aside class="example" id="example-account-security" title="Account Security">
-
-Accounts with a service need to be protected more strongly than just with a username and
-password. Since a compromised account reveals all the private information stored in the
-account, this is a privacy issue and not just a security issue. Services often store the
-historical locations and machine characteristics that have accessed an account, in order
-to make it harder to log in from an unusual place or type of machine. They might also
-store other personal data, like a phone number or address, in order to check that a
-suspicious login actually comes from the real account owner.
-
-</aside>
-
-It is attractive to say that if someone violates the norms of a service they're using,
-then they sacrifice a proportionate amount of their privacy protections, but
-
-1. Often the service can only prevent the norm violation by also collecting data from
-   innocent users. This extra collection is not always [=appropriate=], especially if it
-   allows pervasive monitoring ([[RFC7258]], [[RFC7687]]).
-1. If a service owner wants to collect some extra data, it can be tempting for them to
-   define norms and proportionality that allow them to do so.
-
-<div class="practice">
-
-<p><span class="practicelab" id="principle-limited-collection-for-safety">If a service
-collects extra data from its users in order to protect those or other users, it must take
-extra technical and legal measures to ensure that this data can't be then used for other
-purposes, like to grow the service.</span></p>
-
-This is a special case of the more general principle that data can't be used for more
-[=purposes=] than the data's subjects understood it was being collected for.
-
-</div>
-
-<aside class="example" id="example-technical-legal-measures" title="Technical and Legal Measures">
-
-A site might segregate the data it collects for safety reasons from its business data by:
-
-1. Writing in its privacy policy that these types of data are kept separate. (A legal
-   measure)
-1. Using multi-party computation to ensure that its business side can't learn the
-   sensitive safety data unless both its safety side and a trusted third party collude. (A technical measure)
-1. Hiring a trusted auditor to publicly check that the algorithms are sufficient to
-   provide the guarantees they claim to provide, and that there isn't, in fact, any
-   collusion. (Another technical measure)
-
-</aside>
-
 
 ## Principles for Guardians and Device Owners {#guardians-and-owners}
 
--- a/index.html
+++ b/index.html
@@ -710,19 +710,22 @@ make them more precise.
 
 ## Balancing Different Privacy Principles {#balancing}
 
-Sometimes the most obvious response to one privacy principle is in tension with another
-privacy principle. Often this happens when one user of a service is trying to invade
-another user's privacy: keeping a user's information private can limit the ways that a
-service can protect them or another user.
+While privacy principles are designed to work together and support each other,
+occasionally a proposal to improve how a system follows one privacy principle will reduce
+how well it follows another principle.
 
 <div class="practice">
 
 <p><span class="practicelab" id="principle-pareto-frontier">When confronted with an
-apparent tradeoff, first seek the Pareto frontier.</span></p>
+apparent tradeoff, first look for ways to improve all principles at once.</span></p>
 
-Given any initial design that doesn't perfectly satisfy all goals, there are usually some
-other designs that improve the situation for some goals without sacrificing anything about
-the other goals. Work to find those designs.
+Given any initial design that doesn't perfectly satisfy all principles, there are usually
+some other designs that improve the situation for some principles without sacrificing
+anything about the other principles. Work to find those designs.
+
+Another way to say this is to look for [Pareto
+improvements](https://en.wikipedia.org/wiki/Pareto_efficiency) before starting to trade
+off between principles.
 
 </div>
 
--- a/index.html
+++ b/index.html
@@ -1206,6 +1206,23 @@ categories:
      person trying to create multiple accounts is not having their own safety protected by
      the extra data collection. Instead, other people are being protected from them.
 
+     * A person might want to post false or fraudulent information, or malware, to a
+       content hosting service. While the service can take down this information as it's
+       discovered, if the poster has complete privacy after misbehaving, they can just
+       re-upload the malicious content and can escape prosecution for anything illegal.
+
+     * In a service aimed at children, pedophiles might want to pretend to be children in
+       order to take advantage of the other users. While an ethical platform will likely
+       have content filtering and reporting features, attackers have been known to
+       encourage children to switch to platforms that aren't intended for children in
+       order to avoid these controls.
+
+     * People trying to misinform others will often find it easier when their victims
+       can't find the authors' real affiliations. One aspect of this is
+       [astroturfing](https://en.wikipedia.org/wiki/Astroturfing), where powerful
+       interests pay people or create fake accounts in order to create the impression of
+       agreement from more-sympathetic sources.
+
    * A person might want to pretend to view or click an ad many times, either to convince
      an advertiser to pay the website the ad appeared on, or to cost a particular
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -998,16 +998,90 @@
 
 </div>
 
-## Safety Systems {#safety}
+## Safety, Abuse, and Fraud {#safety}
+
+There are some cases where privacy and safety are in tension. These cases are more rare
```
jbradleychen, line 1003: Concern: These examples under represent the harms and difficulties privacy induces on public safety work. Here are some further examples:

- Privacy of a dishonest publisher facilitates distribution of false information, malware, fraud
- Privacy of a dishonest adult, impersonating a child, enables them to solicit children on a major social media platform, vectoring them into a E2EE environment where criminal acts can occur with military-grade secrecy.
- Privacy-related concerns at major online service providers results in identity services that fail to support users who want to use transparent identity online as a part of legitimate professional activity. It makes it difficult or impossible for users seeking services to differentiate honest professionals from others who are indifferent or hostile to honest identity. The whole of the open Web is harmed.
- Privacy-centric identity systems enable ambiguous identity services, facilitating impersonation of specific individuals (e.g. oprah) and of classes of individuals (americans, BLM activists) facilitating in many forms of fraud and deception.
- PEW research (2014) has reported that about half of online harassment benefits from and is to some degree enabled by anonymity. In other words, weak identity up-to doubles the harassment experienced online. Such harassment disproportionately targets individuals in vulnerable subgroups. 
https://www.pewresearch.org/internet/2014/10/22/online-harassment/

index.html
```diff
@@ -998,16 +998,90 @@
...
+than one might think at first glance, but they still exist. We can distinguish three
```
jbradleychen, line 1004: "more rare than one might think" strikes me as speculative and indefinite. For people who work on safety, privacy presents continuous significant obstacles and destabilizes our efforts.

index.html
philippp: These are valid and also categorically distinct from Jeffrey's list of "tensions," in that they are scenarios in which privacy is weaponized / used to perpetrate harm. Could this accompany Jeffrey's list (e.g. "in addition to these tensions, there are also cases in which privacy becomes a harbor for abuse:")?

index.html
philippp: Agreed, this is highly variable on how much a given reader thinks about it. I'd venture to guess that for the average privacy-minded reader, these trade-offs and abuses are not front of mind (warranting their late inclusion via this PR :])

index.html
```diff
@@ -998,16 +998,90 @@
...
+categories:
```
philippp, line 1005: A person's privacy also rests on implicit safety trade-offs: In order to protect accounts from getting hijacked, services may fingerprint a user's machine and require additional challenges when the account is accessed from a new device. Without stronger protections, retiring this fingerprinting would lead to more account compromises, and the compromise of all private information stored in these accounts.

Would it be possible to accommodate this (or another) example of privacy-enhancing mechanisms themselves requiring "appropriate use of data?"

philippp (commented):
Thanks for the proposal, left a few thoughts about weaponization of privacy and safety trade-offs.

index.html
```diff
@@ -998,16 +998,112 @@
 
 </div>
 
-## Safety Systems {#safety}
+## Safety, Abuse, and Fraud {#safety}
+
+There are some cases where privacy and safety are in tension. Many of these cases are
+short-term, only existing until we develop a system that achieves both privacy and
```
jyasskin, line 1004:
```suggestion
short-term, only existing until the Web community develops a system that achieves both privacy and
```

darobin:
Some notes from the meeting:
* Privacy and safety can be in tension, we need to acknowledge this.
* However, privacy vs safety is a false dichotomy
* Some safety issues are privacy issues (and very much in scope for inclusion in this doc)
* Safety of people matters more than safety of the service
* Safety should not be weaponised against privacy, nor should _removing_ safety be used to force people to accept inferior privacy when alternatives are possible
* Empowering a group with increased rights, power, and information access in order to support safety is a policing decision that we should avoid delegating to opaque and unaccountable entities (eg. a company) but rather should be maintained under expectations of collective governance.

index.html
```diff
@@ -998,16 +998,112 @@
...
+safety. Long-term cases are rare, but still occasionally come up. We can distinguish
```
jbradleychen, line 1005: Can agree that accountability, investigation and auditing are alternatives to data destruction for achieving privacy? If not then the "until we develop a system" language doesn't make sense to me.

Counterabuse attack detection requires long-term access to fairly broad data. Investigation also has data needs that are hard to constrain. As long as privacy is based on destruction of data, I don't see how these safety needs will be met.

Perhaps we need to move beyond the assumption that the only way to implement privacy is to destroy the data. There are many alternative approaches. By comparison, matches are dangerous, but we don't ban them, we use accountability, investigation and auditing mechanisms to make them safe.

index.html
```diff
@@ -998,16 +998,112 @@
...
+three categories:
+
+1. A person's privacy exists in a tradeoff with their own safety.
```
jbradleychen, line 1008: Actually, a publisher's privacy impacts public safety, not just their own safety. Publishers who uses anonymity to avoid accountability for publishing disinformation or to implement fraud cause significant public harm.

index.html
npdoty: Data minimization is an important principle for privacy, but definitely not the only way to support privacy, online or otherwise. My understanding is that there are multiple techniques for supporting user safety as well, not exclusively long-term individual surveillance.

If the document includes text saying the only way to implement privacy is to destroy the data (I don't think that's there, but there are a lot of revisions and I may be behind), that would be a useful separate issue to open.

jyasskin:
I rewrote this [significantly](https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/105.html#balancing), to focus on when different people's privacy rights conflict. In combination with #116, this is clearer that privacy needs to also be balanced with other ethical concerns, including safety, but that this document doesn't give guidelines for doing that other balancing.

I think I'm still missing some text about single-user cases like fingerprinting someone to protect their account from hijacking, which would expose all their information, but I'd be happy to merge this and add the single-user case in another PR.

index.html
jbradleychen: The language here about "short term" and "long term" is inconsistent with my experience. Almost all successful online fraud and malware depends on privacy to be viable. The safety teams I work with spend a substantial part of their time on privacy compliance and on implementing work-arounds for privacy-induced regressions.

The basic constraints for safety are creativity and input data. When enforcement does not have the data to identify individuals with specificity, it leads to collateral damage; false positives or false negatives.

index.html
```diff
@@ -1037,109 +1037,64 @@
...
-## Principles for Safety, Abuse, and Fraud {#safety}
+## Balancing Conflicting Privacy Rights {#balancing}
 
-There are some cases where privacy and safety are in tension. Many of these cases are
-short-term, only existing until we develop a system that achieves both privacy and
-safety. Long-term cases are rare, but still occasionally come up. We can distinguish
-three categories:
+Sometimes different people's privacy rights come into conflict. Often this happens when
+one user of a service is trying to invade another user's privacy: hiding all private
+information about the first user can prevent the service from protecting the second user.
```
jbradleychen, line 1044: Suggestion: "Tension can also arise in cases of fraud and malware, where privacy enables ambiguous or deceptive identity that enables abuse."

index.html
```diff
@@ -1037,109 +1037,64 @@
...
-operating that context may use their data in ways that would otherwise be
-[=inappropriate=] solely in order to protect other users of that service.</span></p>
+<aside class="example" id="example-children" title="Children's Services">
 
-Some extra use of personal data might also be [=appropriate=] in order to protect the
-service itself, but less than is appropriate to protect other people.
+Children using a service (or their guardians) may want to ensure their interaction with
+that service is only visible to other children. This can justify collecting ages from the
+service's users in order to exclude adults, but not to then use those ages for other
+purposes.
```
jbradleychen, line 1087: Yes. Maybe we should be careful not to trivialize age verification. Note that voluntary determination of age is ineffective. Here are some things that could work:
- face recognition, with regular/spontaneous re-verification
- Robust (e.g. [IAL2](https://pages.nist.gov/800-63-3/sp800-63a.html)) verification of an adult guardian, who then attests to the age of their child.

jbradleychen:
Overall this looks much improved. Thanks!

index.html
jyasskin: This text isn't in the PR anymore.

index.html
```diff
@@ -1037,109 +1037,64 @@
...
-     the attacker, the service might still be primarily motivated by preventing damage to
-     itself.
+1. These principles need to be widely-agreed, not just details within a unilateral
+   terms-of-service. For example, in most [=contexts=], it's [=inappropriate=] for a
+   person to misrepresent themself in order to trick other people. On the other hand,
+   there's generally no principle against transparently adopting a pseudonym, even if a
+   site's terms of service prohibit it.
 
-   </aside>
+1. The service must take technical and legal measures to ensure the extra data it's
```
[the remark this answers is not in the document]
jyasskin: What should the consequences be for failing an audit?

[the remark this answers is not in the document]
jbradleychen: Basically I intended the same general "legal measures" suggested by the original text.

More concretely, I imagine a combination of accountability through transparency and subsequent public scrutiny. Transparency means that for a service to be considered in compliance with a hypothetical norm or standard, they would have to participate in a program of transparency and audits. Services that didn't comply/participate would be considered in violation of norms. At one level the consequences would be comparable to other deviations from the principles proposed in this document. Eventually, some regulators might draw from such sources to craft appropriate regulation and consequences.

Just as open source can support an auditing function for source code, transparency can support an auditing function for other situations.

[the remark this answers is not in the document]
dmarti: In general, we can't assume that a web site and a user will have a legal jurisdiction in common. Technical measures, in addition to legal ones, are important in cases where public scrutiny or enforcement action in the user's jurisdiction would have limited or no effect in the site's jurisdiction.

index.html
jyasskin: There was a sense in the [discussion](https://github.com/w3ctag/privacy-principles/blob/main/meetings/2022-01-26-minutes.md#new-topic-105) that I should restrict this section to talking about ways that different privacy interests are in conflict. Otherwise, this document might get too big. Ways that non-privacy principles conflict with our privacy goals should go elsewhere ... although we don't have a good place for them to go.

npdoty: "can prevent" seems like an overstatement. Services can secure data even without successfully collecting all information about a user who tried to invade another's privacy. 

Maybe: Keeping a user's information private may in some cases limit the ways that a service can protect another user.

index.html
```diff
@@ -1037,22 +1037,62 @@
...
+
+<div class="practice">
+
+<p><span class="practicelab" id="principle-violation-sacrifices-privacy">When a
+person violates the [=principles=] of a [=context=] they're interacting with, the service
+operating that context may use their data in ways that would otherwise be
+[=inappropriate=] in order to protect other people.</span></p>
+
+The service may also use innocent people's data in order to distinguish them from these
+bad actors.
```
npdoty, line 1054: I don't think this is intended, but this might read as: bad actors forgo privacy protections by violating the norms of a context and also good actors also forgo those privacy protections if a service may be using that data to protect someone.

That would be a principle that isn't about balance, but that inappropriateness is simply overruled by the purpose of data usage.

index.html
npdoty: I don't believe the proposed text trivializes age verification or that specific technical details are required here. I also don't think our privacy principles document should recommend regular biometric collection and face recognition of children.

index.html
jyasskin: @npdoty We have a couple levels of privacy available here. I wrote "hiding all private information", which is a bit more privacy than your "keeping a user's information private", which is more privacy than "collecting all information about a user". And then those imply a gradient of protection, from my "can prevent" to your "may limit", to more-likely allowing protection.

I think I agree with switching to "may/can limit", since even using all of the first user's plausibly-available information and no specifically-privacy-related restrictions, there's no guarantee that the service will actually be able to protect the second user.

index.html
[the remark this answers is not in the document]
jyasskin: I think @jbradleychen is suggesting something like
```suggestion
1. The service must take technical, legal, and/or social measures to ensure the extra data it's
```

@dmarti is pointing out that purely-legal and purely-social measures are often inadequate, when the site and user are in different jurisdictions, especially when the user's jurisdiction (whether legal or social) is less powerful than the site's.

I'm not confident that public scrutiny is really enough to get sites to behave well, if it's not backed up by regulation or the threat of regulation. But there is legal force behind sites' privacy policies: if they commit to not sharing the data with the growth side of the business, and then an audit discovers they're doing so anyway, I believe regulators do tend to take action. So I think "technical and legal" is sufficient. How's https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/105.html#example-technical-legal-measures to illustrate that?

index.html
jyasskin: We should be intentional about this: if we endorse collecting ages at all, that implicitly gives permission to do it in a way that anti-abuse folks think is effective. If we want to exclude using face recognition on children, we have to say that ... and then we probably shouldn't imply that sites can only use easily-evadable ways of checking ages.

We can always dodge the problem in this document by deleting the example, but it's still going to come up in the real world.

index.html
jyasskin: Good point, I think you've actually read it correctly: the purpose of "preventing users from causing privacy (and other) violations" makes a lot of data collection appropriate (follow the norms/principles of its context) that wouldn't be appropriate for other purposes.

There's still some division between what's appropriate to do with innocent vs guilty users' data: it's probably ok to share guilty users' data across services, while it'd only be ok to use innocent users' data to determine that they're innocent. What would be a better way to say this?

index.html
```diff
@@ -1113,6 +1113,10 @@
 <p><span class="practicelab" id="principle-safety-with-privacy">As much as possible,
 services should provide both safety and privacy.</span></p>
 
+This may require the development and adoption of new technologies. For example, wider
+adoption of [[?WebAuthn]] could reduce account hijacking enough that services no longer
+need to fingerprint devices to prevent it.
+
```
jbradleychen, line 1089: Consider mentioning: 
- However, not all safety problems can be solved through new technologies. Certain systems that combine strong privacy with public safety risk may be impossible to make safe.
Examples I have in mind are WhatsApp with large fan-out, and systems like 8chan that are built around a strategy to prevent accountability.

index.html
jyasskin: https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/105.html#example-account-security now mentions account hijacking.

index.html
jyasskin: https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/105.html#example-children now includes both the pro and con side of collecting ages.

index.html
```diff
@@ -1048,22 +1055,87 @@
...
 
-Discuss what limits need to be placed on such practices to distinguish them from privatised
-law enforcement and intellligence agencies.
+1. Writing in its privacy policy that these types of data are kept separate. (A legal
+   measure)
+1. Using multi-party computation to ensure that its business side can't learn the
+   sensitive safety data unless both its safety side and a trusted third party collude. (A technical measure)
+1. Hiring a trusted auditor to publicly check that the algorithms are sufficient to
+   provide the guarantees they claim to provide, and that there isn't, in fact, any
+   collusion. (Another technical measure)
```
wseltzer, line 1138: Hiring an auditor seems like a policy measure, not "Another technical measure".

jyasskin: Yeah, I wasn't sure how to describe this. It's systematic and operates automatically once it's set in motion, which resembles other technical measures. But it only matters if humans take action based on its outputs, for example by filing lawsuits or exerting social or market pressure. audits are to encryption as [ceremonies](https://eprint.iacr.org/2007/399.pdf) are to protocols? I'm happy to change the wording.

index.html
jyasskin: This text is no longer in the PR.

index.html
```diff
@@ -646,7 +653,89 @@
...
 
+## Balancing Different Privacy Principles {#balancing}
+
+Sometimes different privacy principles come into tension. Often this happens when one
+user of a service is trying to invade another user's privacy: keeping a user's information
+private can limit the ways that a service can protect them or another user.
+
+However, the choice of which privacy principles to prefer is complex and depends heavily
+on the details of each particular situation. <span class="note">Note that people's privacy
+can also be in tension with other important non-privacy concerns. As discussed in the
```
wseltzer, line 664:
```suggestion
However, the choice of which solution to prefer among conflicting responses is complex and depends heavily
on the details of each particular situation. <span class="note">Note that people's privacy
can also be in tension with other important non-privacy concerns. As discussed in the
```

wseltzer: Again, trying to move from "conflicting principles" to conflicting responses.

index.html
```diff
@@ -646,7 +653,89 @@
 Many of the [=principles=] described in the rest of this document extend the [=user agent=]'s duties and
 make them more precise.
 
+## Balancing Different Privacy Principles {#balancing}
+
+Sometimes different privacy principles come into tension. Often this happens when one
```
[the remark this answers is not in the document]
jyasskin: I tried to elaborate on this in 08b71181be30232075e00dae1aabd0aadc8fa765. I think it is actually the principles that are in tension, but we can also say that initial designs are often not the best designs.

index.html
```diff
@@ -653,15 +653,29 @@
...
-discuss how to do that.</span>
+Sometimes the most obvious response to one privacy principle is in tension with another
+privacy principle. Often this happens when one user of a service is trying to invade
+another user's privacy: keeping a user's information private can limit the ways that a
+service can protect them or another user.
+
+<div class="practice">
+
+<p><span class="practicelab" id="principle-pareto-frontier">When confronted with an
+apparent tradeoff, first seek the Pareto frontier.</span></p>
```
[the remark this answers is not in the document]
jyasskin: As soon as there are 2 inalienable rights, you can plot proposals' achievement of each on a diagram and draw a Pareto frontier between them. ... But that may not be obvious from just the term, which invites economic comparisons.

index.html
```diff
@@ -701,8 +708,107 @@
 Many of the [=principles=] described in the rest of this document extend the [=user agent=]'s duties and
 make them more precise.
 
+## Balancing Different Privacy Principles {#balancing}
+
```
sandandsnow, line 712: Consider "Incorporating Different Privacy Principles"

index.html
```diff
@@ -701,8 +708,107 @@
...
+While privacy principles are designed to work together and support each other,
+occasionally a proposal to improve how a system follows one privacy principle will reduce
```
sandandsnow, line 714: Consider ".... follows one privacy principle may reduce ..."

index.html
```diff
@@ -701,8 +708,107 @@
...
+improvements](https://en.wikipedia.org/wiki/Pareto_efficiency) before starting to trade
+off between principles.
+
+</div>
+
+Once one is choosing between different designs at the Pareto frontier, the choice of which
+privacy principles to prefer is complex and depends heavily on the details of each
+particular situation. <span class="note">Note that people's privacy can also be in tension
+with other important non-privacy concerns. As discussed in the [[[Ethical-Web]]], it's
+important to balance these concerns too, but this document doesn't discuss how to do
```
sandandsnow, line 736: Consider "... can also be in tension with other non-privacy concerns. As discussed in the [[[Ethical-Web]]], it's "... important to consider the context in which a particular technology is being applied, the expected audience(s) for the technology, who the technology benefits and who it may disadvantage, and any power dynamics involved..."

(words come from the Ethical Web Principles)

index.html
```diff
@@ -701,8 +708,107 @@
...
+that.</span> Despite this complexity, there is a basic ground rule to follow:
+
+<div class="practice">
+
+<p><span class="practicelab" id="principle-limited-collection-for-safety">If a service
+collects extra data from its users in order to protect those or other users, it must take
```
sandandsnow, line 742: Consider "If a service needs to collect extra data ..."

index.html
```diff
@@ -701,8 +708,107 @@
...
+</div>
+
+<aside class="example" id="example-technical-legal-measures" title="Technical and Legal Measures">
+
+A site might segregate the data it collects for safety reasons from its business data by:
+
+1. Writing in its privacy policy that these types of data are kept separate. (A legal
+   measure)
+1. Using multi-party computation to ensure that its business side can't learn the
+   sensitive safety data unless both its safety side and a trusted third party collude. (A technical measure)
```
sandandsnow, line 758: Consider "a trusted independent third party" (to emphasize independence and hopefully less risk of collusion)

index.html
```diff
@@ -701,8 +708,107 @@
...
+This is a special case of the more general principle that data can't be used for more
+[=purposes=] than the data's subjects understood it was being collected for.
+
+</div>
+
+<aside class="example" id="example-technical-legal-measures" title="Technical and Legal Measures">
+
+A site might segregate the data it collects for safety reasons from its business data by:
+
+1. Writing in its privacy policy that these types of data are kept separate. (A legal
```
sandandsnow, line 755: Consider "1. Specifying in its privacy statement that these types of data are kept separate and implementing policies and procedures to ensure the data is stored separately. (legal and procedural/compliance measures)"

index.html
```diff
@@ -701,8 +708,107 @@
...
+   measure)
+1. Using multi-party computation to ensure that its business side can't learn the
+   sensitive safety data unless both its safety side and a trusted third party collude. (A technical measure)
+1. Hiring a trusted auditor to publicly check that the algorithms are sufficient to
+   provide the guarantees they claim to provide, and that there isn't, in fact, any
+   collusion. (Another technical measure)
+
```
sandandsnow, line 762: Consider also "Hiring a trusted auditor to publicly check that the data is stored separately (a compliance measure)"

index.html
```diff
@@ -701,8 +708,107 @@
...
+On the other hand, identifying everyone with enough detail to detect these cases tends to
+violate their rights to be free from [=surveillance=] and [=correlation=].
+
+</aside>
+
+<aside class="example" id="example-children" title="Children's Services">
+
+Children using a service (or their guardians) may want to ensure their interaction with
+that service is only visible to other children. To accomplish this, the service would need
+to verify all of its users' ages.
```
sandandsnow, line 793: Consider "To accomplish this, services may apply different approaches such as: a service may ask users to self-assert that they are under the specified age (information that is not verified), or the service may only be made available to a trusted institution with knowledge of the users' ages (age verified by a trusted third party and not disclosed to the service, such as a school), or the service may ask its users to prove that they are under a specified age (e.g., by presenting some form of identification, with or without biometrics) (the most privacy-invasive approach).

index.html
```diff
@@ -701,8 +708,107 @@
...
+
+</aside>
+
+It is attractive to say that if someone violates the norms of a service they're using,
+then they sacrifice a proportionate amount of their privacy protections, but
+
+1. Often the service can only prevent the norm violation by also collecting data from
+   innocent users. This extra collection is not always [=appropriate=], especially if it
+   allows pervasive monitoring ([[RFC7258]], [[RFC7687]]).
+1. If a service owner wants to collect some extra data, it can be tempting for them to
```
sandandsnow, line 771: Consider something along these lines "Services should specify what data they collect and process to prevent breaches of terms of service or other norms violations and what data they may collect and process in the event of a breach or violation."

index.html
jyasskin: Done. I'm not sure I exactly like this new heading, but I'd rather get this merged than wait for a perfect option.

index.html

index.html

index.html

index.html
```diff
@@ -701,8 +708,107 @@
...
+<div class="practice">
+
+<p><span class="practicelab" id="principle-limited-collection-for-safety">If a service
+collects extra data from its users in order to protect those or other users, it must take
+extra technical and legal measures to ensure that this data can't be then used for other
+purposes, like to grow the service.</span></p>
+
+This is a special case of the more general principle that data can't be used for more
+[=purposes=] than the data's subjects understood it was being collected for.
+
```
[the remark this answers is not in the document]
jyasskin: I left the last third of the sentence after "purposes", but I expect we'll amend this paragraph to link to the more general principle once we've written it.

index.html

index.html

index.html
jyasskin: I used "effectively segregated" to emphasize that the data has to be kept apart even while processing it.

index.html
jyasskin: Should we have a general principle about telling people what data is collected? I added something about this inside the principle above, with somewhat simpler language.

index.html
jyasskin: I reworded this to blend into the next paragraph.

jyasskin (commented):
Thanks!

PR #114, opened 2022-01-26
Add section on principles for vulnerable people

darobin:
***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/pull/114.html#principles-for-vulnerable-people" title="Last updated on Feb 2, 2022, 5:01 PM UTC (1254623)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/114/3b295f1...1254623.html" title="Last updated on Feb 2, 2022, 5:01 PM UTC (1254623)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1094,8 +1094,7 @@ protected by the [=user agent=] as well as by [=parties=].
 One issue, however, is that signalling to [=parties=] that a [=person=] may be [=vulnerable=] is
 itself a sensitive piece of information that should not be shared. One way to approach this issue is
 to rely on a [=global opt-out=] mechanism such as [[?GPC]] that is used by [=people=] whether they
-are [=vulnerable=] or not and puts a requirement on [=parties=] to minimise their [=data=]
-[=processing=].
+are [=vulnerable=] or not.
 
 <div class="practice">
 
--- a/index.html
+++ b/index.html
@@ -1099,7 +1099,7 @@ are [=vulnerable=] or not and puts a requirement on [=parties=] to minimise thei
 
 <div class="practice">
 
-<span class="practicelab" id="principle-identity-per-context">[=User agents=] should support a mode
+<span class="practicelab" id="principle-protecting-vulnerable">[=User agents=] should support a mode
 to specifically protect [=vulnerable=] people that defaults to sending a [=global opt-out=] signal
 such as [[?GPC]].</span>
 
--- a/index.html
+++ b/index.html
@@ -1086,8 +1086,8 @@ law enforcement and intellligence agencies.
 
 ## Principles for Vulnerable People
 
-Some people, or groups of people, are at greater risk from the asymmetries of power created by the
-concentration of information power ([[?Automating-Inequality]]). [=People=] who for whatever reason
+Some people, or groups of people, are more easily coerced into giving up their privacy.
+[=People=] who for whatever reason
 are in a more [=vulnerable=] position with respect to data [=processing=] should be more proactively
 protected by the [=user agent=] as well as by [=parties=].
 
--- a/index.html
+++ b/index.html
@@ -1086,9 +1086,25 @@ law enforcement and intellligence agencies.
 
 ## Principles for Vulnerable People
 
-<aside class="issue">
-  This section has not been written.
-</aside>
+Some people, or groups of people, are at greater risk from the asymmetries of power created by the
+concentration of information power ([[?Automating-Inequality]]). [=People=] who for whatever reason
+are in a more [=vulnerable=] position with respect to data [=processing=] should be more proactively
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1086,9 +1086,25 @@
 
 ## Principles for Vulnerable People
 
-<aside class="issue">
-  This section has not been written.
-</aside>
+Some people, or groups of people, are at greater risk from the asymmetries of power created by the
+concentration of information power ([[?Automating-Inequality]]). [=People=] who for whatever reason
```
jyasskin, line 1090: This can probably be simpler ... maybe

```suggestion
Some people, or groups of people, are more easily coerced into giving up their privacy.
[=People=] who for whatever reason
```

index.html
```diff
@@ -1086,9 +1086,25 @@
 
 ## Principles for Vulnerable People
 
-<aside class="issue">
-  This section has not been written.
-</aside>
+Some people, or groups of people, are at greater risk from the asymmetries of power created by the
```
jyasskin, line 1089: Do we need both "people" and "groups of people"? Can a group be at risk without its individual members being at risk?

In the "collective issues" section, it makes sense to talk about groups as separate from individuals, but I think the reasons there don't apply here.

index.html
```diff
@@ -1086,9 +1086,25 @@
...
+itself a sensitive piece of information that should not be shared. One way to approach this issue is
+to rely on a [=global opt-out=] mechanism such as [[?GPC]] that is used by [=people=] whether they
+are [=vulnerable=] or not and puts a requirement on [=parties=] to minimise their [=data=]
+[=processing=].
+
+<div class="practice">
+
+<span class="practicelab" id="principle-identity-per-context">[=User agents=] should support a mode
+to specifically protect [=vulnerable=] people that defaults to sending a [=global opt-out=] signal
+such as [[?GPC]].</span>
```
jyasskin, line 1104: Do we have a reason to think that vulnerable people will be able to send the global opt-out signal? If they can be coerced away from sending individual opt-out signals, it seems like that coercion will still apply in the context of the global one.

index.html
```diff
@@ -1086,9 +1086,25 @@
 
 ## Principles for Vulnerable People
 
-<aside class="issue">
-  This section has not been written.
-</aside>
+Some people, or groups of people, are at greater risk from the asymmetries of power created by the
+concentration of information power ([[?Automating-Inequality]]). [=People=] who for whatever reason
+are in a more [=vulnerable=] position with respect to data [=processing=] should be more proactively
+protected by the [=user agent=] as well as by [=parties=].
```
jyasskin, line 1092: Given the caveat below that it's generally inappropriate to identify (or try to identify?) that a person is vulnerable, I'm not sure how we can say to proactively protect that specific group of people.

Maybe it's that "[=Actors=] on the web should not take advantage of any power they have over vulnerable people and should strive to protect any vulnerable people they happen to identify from other actors."? (This could be a Principle.) And then maybe something about actors needing to get their designs reviewed with user advocates and other experts to make sure they're not accidentally doing so?

index.html
```diff
@@ -1086,9 +1086,25 @@
...
+
+One issue, however, is that signalling to [=parties=] that a [=person=] may be [=vulnerable=] is
+itself a sensitive piece of information that should not be shared. One way to approach this issue is
+to rely on a [=global opt-out=] mechanism such as [[?GPC]] that is used by [=people=] whether they
+are [=vulnerable=] or not and puts a requirement on [=parties=] to minimise their [=data=]
+[=processing=].
```
jyasskin, line 1098: I think parties already have a requirement to minimise their data processing ... Maybe just remove this clause:

```suggestion
are [=vulnerable=] or not.
```

The point here is to let the vulnerable people blend into a larger population, not to re-describe what GPC in particular does.

jyasskin (changes requested):

index.html
darobin: I would rather keep it as this can apply for instance to cohorts or to associations (say, data collected about unions or shelters).

index.html
jbradleychen: In addition to these safety mechanisms, process transparency and auditability seem relevant, not just regarding services for vulnerable people but also in other situations too where risks associated with data flows cannot be simply eliminated.

We might also be mindful about creating new obstacles to creators of services for vulnerable people. Assuming we ultimately want more and better services for vulnerable people, making such services extremely difficult to implement could have unintended consequences.

wseltzer:
I'm not sure what you're trying to do with this section. Is it to justify having and respecting global signals, so that vulnerable populations can hide in larger crowds without disclosing their unique characteristics?

PR #122, opened 2022-02-04
include warning about how principles aren't done

darobin:
During the last call we discussed the fact that we could message to readers the idea that the Principles section wasn't as baked but that that shouldn't be read as indicating that we don't care about one thing or another. If we do go ahead, I'd suggest something like the below.

### DIFF
--- a/index.html
+++ b/index.html
@@ -661,6 +661,12 @@ is that a [=context=] is not defined in terms of who owns it (it is not a [=part
 [=data=] between different [=contexts=] of a single company is just as much a [=privacy violation=] as
 if the same data were shared between unrelated [=parties=].
 
+<aside class="issue">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -661,6 +661,12 @@
 [=data=] between different [=contexts=] of a single company is just as much a [=privacy violation=] as
 if the same data were shared between unrelated [=parties=].
 
+<aside class="issue">
+  Please note that the initial focus of this draft was on <a href="#intro"></a>. This section is
+  incomplete and less mature. The absence of a specific principle should not be seen as an
```
jyasskin, line 666: Maybe

```suggestion
  incomplete and has uneven maturity. The absence of a specific principle should not be seen as an
```

jyasskin (approved):

PR #143, opened 2022-04-14
Wording on control of information

dmarti:
As discussed in today's meeting -- make it clear that the first party
is the party that has control of the person's (user's) data.  (The
editor of site content or the administrator of the site's hosting
service is not necessarily the first party.)

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/pull/143.html" title="Last updated on Apr 21, 2022, 5:15 PM UTC (74e723e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/143/ff66160...74e723e.html" title="Last updated on Apr 21, 2022, 5:15 PM UTC (74e723e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1554,7 +1554,7 @@ the core concepts of that domain combine into a [=user=]-comprehensible [=party=
 definitions are likely to differ between domains.
 
 [=User agents=] tend to explain to [=people=] which [=origin=] or [=site=] provided the
-web page they're looking at. The [=party=] that controls the use of information collected on an [=origin=] or [=site=] is
+web page they're looking at. The [=party=] that controls this [=origin=] or [=site=] is
 known as the web page's <dfn data-dfn-for="page">first party</dfn>. When a [=person=]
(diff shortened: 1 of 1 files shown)


jyasskin (approved):
There may be some nuance here for our principles around interruptions and misinformation, but I think we're mostly using first-party-ness for information use anyway, so I think this change will be fine.

index.html
```diff
@@ -1554,12 +1554,13 @@
 definitions are likely to differ between domains.
 
 [=User agents=] tend to explain to [=people=] which [=origin=] or [=site=] provided the
-web page they're looking at. The [=party=] that controls this [=origin=] or [=site=] is
+web page they're looking at. The [=party=] that controls the use of information collected on an [=origin=] or [=site=] is
```
darobin, line 1557: I don't think that this is correct? The use of the information collected on an origin is often controlled by third parties. In fact, that's a key problem!

Issue #150, opened 2022-04-20
when is sharing data for some ancillary purpose consistent with privacy principles?


npdoty:
There seem to be open questions about when it might be acceptable and consistent with a privacy threat model for the Web for some data to be shared about a user's actions that isn't directly necessary for the user's current action. 

This might be to enable some generally beneficial ancillary functionality, or for harm reduction purposes (alternatives to more invasive data collection that will happen otherwise), or where it's just not feasible to minimize data in the short/medium term.

Examples could include:
* Reporting APIs, Beacon, performance measurements, debugging/troubleshooting
* statistics on links clicked or advertising attribution
* `User-Agent` string/hints used for analytics

This discussion was prompted in part by #133 (introducing a principle on not sharing data unless necessary for overt immediate goals) and in part by text around collective privacy/governance. #121 may also be an example.

yoavweiss:
Related: https://github.com/w3ctag/design-principles/issues/368

yoavweiss:
I think we need to make a clear distinction between ancillary data that is not available by other means, and ancillary data that is 1:1 equivalent with other data that's available to the web site for functional purposes. The former is purely ancillary, while the latter is somewhat different.

npdoty:
Ancillary uses of data is somewhat independent of what data is accessed (that is, data itself isn't ancillary or not).

#182 is proposed for handling cases where data is otherwise available (either currently or indefinitely).

torgo:
We're closing this as we believe we addressed this in #184 and #182.

npdoty:
https://github.com/w3ctag/privacy-principles/pull/216 is proposed to revise answers to this question.

torgo:
Agreed to close due to imminent merger of https://github.com/w3ctag/privacy-principles/pull/216.

Issue #106, opened 2022-01-23
Privacy and Safety as joint goals


jbradleychen:
Regarding https://github.com/w3ctag/privacy-principles/pull/102 :

My main overall concern with this revision and draft is that it tends to cherry-pick privacy out of the principles for an <a href="https://www.w3.org/2001/tag/doc/ethical-web-principles/">ethical web</a>, pursuing that goal in isolation but in the process ignoring big problems that result for other co-equal priorities. Some specific principles from the <a href="https://www.w3.org/2001/tag/doc/ethical-web-principles/">ETHICAL-WEB</a> draft:
- _2.2 The Web should not cause harm to society._ This draft largely ignores ongoing societal harms of fraud, malware, ransomware, disinformation, and exploitation of vulnerable users, and the impact of privacy in facilitating these harms.
- _2.5 Security and privacy are essential._ This draft largely ignores the public safety aspects of security, in conflict with this principle.
- _2.7 The web must make it possible for people to verify the information they see._ This draft does not consider the essential role of honest, stable identity in the most ordinary information verification tasks. This omission is problematic and fundamental given the relationship between identity transparency and privacy. 
- _2.8 The web must enhance individuals' control and power._ Information is power. To provide meaningful user control and power, users must have the information to make good trust decisions online. In the absence of verifiable context information, users will make decisions based on unverifiable information, leading in the (very common) degenerate case to random information and random decisions, effectively vacating individual control and power. 

In order for this document to usefully advance the ethical web, contributors need to be mindful of and avoid privacy absolutism, and explicitly consider the need for balance between privacy and the public safety considerations with which privacy is commonly in direct opposition. The sad truth is that perfect privacy provides perfect cover for criminal activity, and if privacy proponents do not want to further facilitate such unsafe conditions, they need to recognize and embrace the difference between perfect privacy and privacy in balance. Given the state of the systems in question, I assert this is only possible if excellent privacy and excellent public safety are considered together, simultaneously, as requirements of the same solution, and not by cherry-picking privacy out of the <a href="https://www.w3.org/2001/tag/doc/ethical-web-principles/">ETHICAL-WEB</a> draft.

[...]

PR #158, opened 2022-05-10
Tension within anti-profiling designs

jyasskin:
***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/158.html#example-preventing-profiling" title="Last updated on Jan 4, 2023, 5:59 PM UTC (fe3db6f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/158/0b159b4...jyasskin:fe3db6f.html" title="Last updated on Jan 4, 2023, 5:59 PM UTC (fe3db6f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -964,10 +964,10 @@ principles.
 User agents can reduce the incentive to develop these alternate tracking methods by
 building APIs to facilitate the most common and least harmful uses of user profiles.
 However, those APIs usually still reveal some information about a user's behavior. For
-example, [[[IPA]]] reveals a limited amount of information about each user in order to
-replace the use of profiles to measure the success of advertising campaigns. However, even
-a small amount of information is still personal data, which the principles in [[[#data]]]
-advise against sharing.
+example, even the most privacy-respecting conversion attribution API will reveal a limited
+amount of information about each user in order to replace the use of profiles to measure
+the success of advertising campaigns. However, even a small amount of information is still
+personal data, which the principles in [[[#data]]] advise against sharing.
 
 </aside>
 
--- a/index.html
+++ b/index.html
@@ -966,8 +966,8 @@ building APIs to facilitate the most common and least harmful uses of user profi
 However, those APIs usually still reveal some information about a user's behavior. For
 example, even the most privacy-respecting conversion attribution API will reveal a limited
 amount of information about each user in order to replace the use of profiles to measure
-the success of advertising campaigns. However, even a small amount of information is still
-personal data, which the principles in [[[#data]]] advise against sharing.
+the success of advertising campaigns. Even this small amount of information is still
+[=personal data=].
 
 </aside>
 
--- a/index.html
+++ b/index.html
@@ -953,22 +953,21 @@ suspicious login actually comes from the real account owner.
 
 <aside class="example" id="example-preventing-profiling" title="Preventing Profiling">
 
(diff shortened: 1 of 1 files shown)


yoavweiss:
Should this also include the possibility of applications moving off of the web and onto other application platforms that are more privacy invasive? Or does this belong in some other document?

npdoty:
Is this an example of recommendations from different principles that are in tension? I don't know what the two different groups of principles are that the text refers to.

It seems more like an example of how mitigations are imperfect or how harm reduction can be used to mitigate but not eliminate privacy impacts.

npdoty:
@jyasskin the updated text helps me understand the suggestion, thank you!

I don't particularly see recommendations being in tension here -- user agents should help users select the identity they want to present and should share the data that the user wants to share (and not the data they don't want to share). Maybe this will depend in part on how we nail down the principle about what data about a user is shared (e.g. #150).

I don't know that providing alternative targeting mechanisms is done for the purpose of, or is likely to have the effect, of largely decreasing the funding available to develop alternative tracking mechanisms, unless this alternative targeting mechanism is so effective and so widely used that additional surveillance could not have any marginal effect for any party -- a very high bar.

npdoty:
I agree that "moving off the web" is a potential risk to consider. I'm not sure how exactly that would lead us to different conclusions for privacy on the Web (which is what I would consider the scope of the document), but I'd be open to pointers or thinking about that.

I'm certainly interested in ways to advocate and design for better privacy on other platforms as well. I think it's unlikely that an effective way to improve privacy on other platforms is to make sure we don't have too much privacy for users of the Web. :)

jyasskin:
@wseltzer We'd agreed to kick tradeoffs between privacy and non-privacy concerns out to another document, but this one is between two different privacy goals, and I don't see another document that it would fit in. It's true that this is less direct than the other tradeoffs, but it's really important to consider and plan for these second-order effects when changing how the Web works.

wseltzer:
@jyasskin I agree second-order effects are important, but I'm less convinced we can predict them confidently, nor that we can then enumerate subsequent follow-on effects including unintended consequences. I think it's a difference between general principle and implementation detail, where successful implementation depends on iteration and empirical testing.

torgo:
cross-referencing #106

index.html
```diff
@@ -837,6 +843,26 @@
...
+[[[#identity]]] by blocking this profiling, but because the profiles are valuable, there's
+a large incentive to work around user agent measures, sometimes by using techniques that
+are very expensive or impossible to block. If user agent behavior causes websites to adopt
+these alternate tracking methods, the web as a whole won't respect the [[[#identity]]]
+principles.
+
+User agents can reduce the incentive to develop these alternate tracking methods by
+building APIs to facilitate the most common and least harmful uses of user profiles.
+However, those APIs usually still reveal some information about a user's behavior. For
+example, [[[IPA]]] reveals a limited amount of information about each user in order to
```
jyasskin, line 859:
```suggestion
example, even the most privacy-respecting conversion attribution API will reveal a limited amount of information about each user in order to
```

jyasskin:
Merging as discussed on [Jan 4](https://github.com/w3ctag/privacy-principles/blob/main/meetings/2023-01-04-minutes.md#pr-tension-within-anti-profiling-designs).

PR #162, opened 2022-05-11
Allow non-immediate goals in data sharing

jyasskin:
And use wording from a pre-existing document instead of drafting our own. This other wording would cover both sites and browsers, where our wording only covered browsers.

I tweaked their wording to switch user->person and she->they.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/162.html#data" title="Last updated on May 18, 2022, 4:38 PM UTC (0bd114b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/162/7c922f1...jyasskin:0bd114b.html" title="Last updated on May 18, 2022, 4:38 PM UTC (0bd114b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -172,6 +172,13 @@
           href: 'https://www.cambridge.org/core/books/governing-privacy-in-knowledge-commons/FA569455669E2CECA25DF0244C62C1A1',
           publisher: 'Cambridge University Press',
         },
+        'Governance-Through-Privacy': {
+          title: "Governance Through Privacy, Fairness, and Respect for Individuals",
+          authors: ['Dixie B. Baker', 'Jane Kaye', 'Sharon F. Terry'],
+          date: '2016-03-31',
+          publisher: "eGEMs",
+          href: 'https://doi.org/10.13063/2327-9214.1207',
+        },
         'GPC': {
           title: 'Global Privacy Control (GPC)',
           authors: ['Robin Berjon', 'Sebastian Zimmeck', 'Ashkan Soltani', 'David Harbage', 'Peter Snyder'],
@@ -1050,12 +1057,14 @@ of an event that occurs nearly-simultaneously on both sites.
 <div class="practice">
 
 <p><span class="practicelab" id="user-agent-data-sharing">
-User agents should only share and present information about the user that
-the user can reasonably anticipate being shared, and which directly relate to the users overt, immediate goals.
+A person's personal information should never be collected, used, transmitted, or disclosed
+in a way that would surprise the person were they to learn about it
+[[Governance-Through-Privacy]]. This information should also be needed for the person's
+goals.</span></p>
+
 User agents and sites should be conservative when deciding if information is related to a users goals; the
 longer the chain of reasoning used to justify data collection, the less likely for that data
 collection to be ethical and principled.
-</span></p>
 
 </div>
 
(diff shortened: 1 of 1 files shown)


pes10k:
I dont think we should adopt or hook into HIPAA-documents for the W3C document. The domains are very different, health care data is probably more predictable than browser-accessible data in many ways (and probably less so in others), and in general HIPAA doesn't seem to me to me to be a clearly good model to build on.

More importantly, I don't think "no surprises" is a good line to try and draw a principal around.  One, it doesn't seem to provide any protection for the many folks in the "i know my device is harming my privacy, but what can I do" category, or "privacy defeatism" in general.  It also seems like an unworkable principal (who's going to judge whats surprising to users we don't know and never will, and who have technical expertise widely different than ours).

In general, I think better to keep the focus on "only collect data that immediately serves what the user is trying to accomplish, and ask when thats not clear." Its difficult enough for sites to judge user goals; I expect it's impossible for sites to know what kinds of data collection would be surprising.

samuelweiler:
I like the affirmative wording in the original version.  I'm not averse to adding the "no surprise" language also, but I'd like to also keep the original, affirmative wording.

index.html
```diff
@@ -1050,12 +1050,12 @@
 <div class="practice">
 
 <p><span class="practicelab" id="user-agent-data-sharing">
-User agents should only share and present information about the user that
-the user can reasonably anticipate being shared, and which directly relate to the users overt, immediate goals.
+User agents should only share information about the user that the user can reasonably
+anticipate being shared, and which directly relate to the user's goals.</span></p>
```
jyasskin, line 1054: @dmarti suggested that we think about the user's goals in terms of what they would do themself if they had enough time. So something like

```suggestion
anticipate being shared, and which the user would do themself if they had complete information and enough time.</span></p>
```

or s/enough time/could do as many things each second as their computer can/. There's still something missing about how the UA guesses what the user would do themself: is that what a reasonable observer would guess the user would do themself, or is the "should" enough to capture that any agent has to guess about their principal's wishes?

@hober was worried that even given unlimited time, the user might not bother (her example) reading the T&Cs or (my example) telling an advertiser where they saw the ad that got them to buy a thing because it's more fun to just go watch a movie. That's probably fine ... if they don't actually have enough of a goal to fund their websites that they'd dedicate an insignificant amount of time to it, it's reasonable to think their UA shouldn't be doing it either.

jyasskin, line 1054: Another simpler change here might be 
```suggestion
anticipate being shared, and which are needed for the user's goals.</span></p>
```
I've disliked "relate to" elsewhere for being vague, and I was overruled in those cases, but maybe this one will stick.

dmarti: People who make well-informed decisions to buy legitimate products are fairly likely to answer reasonable questions from the seller about "where did you see our ad?" But most people would not log and report on all the ads they see, good and bad.

index.html
```diff
@@ -1050,12 +1050,12 @@
...
+
 User agents and sites should be conservative when deciding if information is related to a users goals; the
 longer the chain of reasoning used to justify data collection, the less likely for that data
 collection to be ethical and principled.
-</span></p>
```
hober, line 1058: Line 1052 has `<p><span>` so I assume we shouldn't drop the closing tags here?

jyasskin:
I believe the text this is changing is no longer in the document.

PR #166, opened 2022-05-23
add additional details to clarify privacy by default principle

### LINKED ISSUE #163: Principle 2.1 (Identity on the Web) should account for users not wishing to share their identity
Looking at the way the principle is defined currently creates a good foundation for accurately reflecting how identity on the web works today. However, it doesn't accurately reflect many users (myself included) on the web goal where if I choose to not share any aspect of my identity I should be able to do so. I think this principle should be updated to account for when the user chooses not to share any information even in a first party context. While this isn't necessarily perfectly possible on the web today (e.g. any time you make a web request you're probably sharing an IP address), I think it's a good north star that would be a good thing to include as a foundational principle on the web that we aim to achieve.

To reconcile this position with the common tension that web services typically require additional information in order to legitimately provide their service (e.g. needing location information for regulatory requirements) we should also account for the fact that the user is entering in a negotiation of sorts where they're trading their information in exchange for access and usage of services on the web. This should be done through a consent based approach that shouldn't assume that bundles of attributes (e.g. the identity of the user) need to all be shared at the same time and instead the information can be gradually shared as the user wants to grant more information to the service. 

This model of information sharing most accurately reflects what we as people are used to in meatspace where walking in a public space I generally assume that a researcher sitting in the public space can gather information about my physical attributes like what I'm wearing or can take a photo of me. However, the researcher must ask me for each one of my attributes such as name or date of birth in order to gather that information. They can't inherently infer this easily just based on my appearance (let's leave the celebrity edge case out of this analogy for now). This is similar to on the web the web service can gather information about my location based on my IP address today, but the user can make attempts to hide it such as via a proxy or VPN service. The difference is that on the web we have a greater granularity of the information we reveal as we access services on the web and therefore have a greater ability to reveal less unless we (as users) explicitly consent to sharing this information in exchange for a greater amount of access to services on the web.

kdenhartog:
closes #163

***
<a href="https://pr-preview.s3.amazonaws.com/kdenhartog/privacy-principles/pull/166.html" title="Last updated on Nov 4, 2022, 12:35 AM UTC (4212c89)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/166/90459b8...kdenhartog:4212c89.html" title="Last updated on Nov 4, 2022, 12:35 AM UTC (4212c89)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -922,9 +922,11 @@ site.
 
 A [=person=]'s <dfn>identity</dfn> is the set of characteristics that define
 them. Their identity *in a [=context=]* is the set of characteristics they
-present in that context. People frequently present different identities to
-different contexts, and also frequently share an identity among several
-contexts.
+present to that context, or that the context can infer.  This set of
+characteristics can be intentionally small and can differ from the person's
+physical-world characteristics or [=identifiers=]. People frequently present
+different identities to different contexts, and also frequently share an
+identity among several contexts.
 
 <dfn>Cross-context recognition</dfn> is the act of recognising that an
 [=identity=] in one [=context=] is the same [=person=] as an [=identity=] in


kdenhartog:
cc @jyasskin since it didn't let me add you as a reviewer in the github tooling

index.html
```diff
@@ -918,6 +921,10 @@
 people need. For example, some user agents may help people present different
 [=identities=] to subdivisions of a single [=site=].
 
+In each of these [=contexts=] the UA should assume by default that the user
```
kdenhartog, line 924: I was uncertain if this text should be elevated to the principle text box because I hadn't spent enough time looking into what should and should not go within there. I put it below since the text right above this talks about defaults, but I'd be happy to see this text go within the principle text box if others also believe it warrants that level of attention.

index.html
```diff
@@ -918,6 +921,10 @@
...
+wishes to present an [=identity=] without any identifiable characteristics to
+remain as private as possible unless configured or constented to explicitly by the user.
```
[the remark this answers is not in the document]
kdenhartog: addressed in: https://github.com/w3ctag/privacy-principles/commit/4bfb775551e7c70d8ba86e7d43088001c44d004d

Thanks for the suggestion

sandandsnow:
Thanks for the helpful input @kdenhartog. I'm going to propose some simplification, namely:

- "Their identity in a context is the set of characteristics they present to that context, directly or indirectly."
- "In certain cases, people may also choose to present no characteristics or false characteristics about themselves."

Also, good addition of "behavior".

As for "browser identity", I think you might mean "browser fingerprint", which would need a separate bullet point.

index.html
```diff
@@ -864,9 +864,12 @@
...
 A [=person=]'s <dfn>identity</dfn> is the set of characteristics that define them. Their identity
-*in a [=context=]* is the set of characteristics they present to that context. People
-frequently present different identities to different contexts, and also frequently share an identity
-among several contexts.
+*in a [=context=]* is the set of characteristics that can be identified about
+them either through direct presentation or other means. People frequently
+present different identities to different contexts, and also frequently share an
+identity among several contexts. In certain cases, people may also choose to present no
+characteristics or false characteristics about themselves to maintain a more
+private identity in certain [=context=].
```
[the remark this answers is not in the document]
kdenhartog: Thank you @samuelweiler 

addressed this in [`0541180` (#166)](https://github.com/w3ctag/privacy-principles/pull/166/commits/054118007014a4bd9c11ba6ad1e6c0eae2c909d6)

kdenhartog:
> Thanks for the helpful input @kdenhartog. I'm going to propose some simplification, namely:
> 
> * "Their identity in a context is the set of characteristics they present to that context, directly or indirectly."
> * "In certain cases, people may also choose to present no characteristics or false characteristics about themselves."
> 
> Also, good addition of "behavior".
> 
> As for "browser identity", I think you might mean "browser fingerprint", which would need a separate bullet point.

I've attempted to address these comments.

Also for the browser fingerprinting that's actually being changed in #167 and this is a PR that I rebased on top of that since I was uncertain if that would be accepted so wanted to create a separate discussion, but because I built this branch on top of that one, it ended up combining these all together.

We can either merge #167 first and this second to avoid conflicts or close it and discuss it all on this thread. Either works for me.

kdenhartog:
I've rebased `main` onto here which now included #167 text that was merged so that this PR can stay focused on data minimization concerns to minimize an identity on the web by default. I'll take a look at the notes from today as well once they're posted to see if I can get a read for where others in the group are at and what changes still need to be reflected in this PR to get it merged.

Thanks for the review and discussion so far on it!

hadleybeeman:
look

kdenhartog:
I think I figured out what was being intended, just got a bit confused when I originally was reading through stuff. I've removed the second paragraph which can be discussed in #171.

I made modifications to the original text in https://github.com/w3ctag/privacy-principles/pull/166/commits/13e5c921e70a6a3442f096483facee04ed102742 so I think this one is ready to merge now. 

Wdyt?

index.html
```diff
@@ -864,9 +864,12 @@
...
 A [=person=]'s <dfn>identity</dfn> is the set of characteristics that define them. Their identity
-*in a [=context=]* is the set of characteristics they present to that context. People
-frequently present different identities to different contexts, and also frequently share an identity
-among several contexts.
+*in a [=context=]* is the set of characteristics they present to that context,
+directly or indirectly. People frequently present different identities to
+different contexts, and also frequently share an identity among several
+contexts. In certain cases, people may also choose to present minimal unique
+characteristics or false characteristics about themselves in certain
+[=contexts=] to achieve greater privacy on the web.
```
jyasskin, line 872: How about this rewording? It groups the descriptions of the characteristics together, instead of putting the description of identity sharing and non-sharing in the middle.

```suggestion
*in a [=context=]* is the set of characteristics they present to that context,
directly or indirectly. This set of characteristics can be intentionally small
and can differ from the person's physical-world characteristics. People
frequently present different identities to
different contexts, and also frequently share an identity among several
contexts.
```

index.html
```diff
@@ -864,9 +864,12 @@
 </aside>
 
 A [=person=]'s <dfn>identity</dfn> is the set of characteristics that define them. Their identity
-*in a [=context=]* is the set of characteristics they present to that context. People
-frequently present different identities to different contexts, and also frequently share an identity
-among several contexts.
+*in a [=context=]* is the set of characteristics they present to that context,
+directly or indirectly. People frequently present different identities to
```
jyasskin, line 868: "directly or indirectly" shows up in a couple places, and I'm never sure exactly what it means. Here, is it
```suggestion
*in a [=context=]* is the set of characteristics they present to that context,
or that the context can infer. People frequently present different identities to
```
?

jyasskin (commented):
Sorry I never reviewed this. The idea seems solid.

index.html
kdenhartog: I think it does a good job of highlighting some of the key points, but it does leave out that from a web pages perspective the identity of the user is actually the combination of the user's attributes and the UA's attributes. If we focus the concept of an identity too much on the user it seems like we lose the value of calling out that the UA is actually what gets used to uniquely identify users most of the time. I'd like to make that more understood in this text so that we're leveraging these principles to inform ourselves on better design choices in the future.

For example, if I configure my UA to behave as the example describes in [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#section-12.5.4) then that's describing a preference of the UA not me as I can understand `en-gb` the exact same as `en`, but this attribute could be significantly unique enough to fingerprint the UA, and by extension me the user if I'm the only user who uses this browser.

A way to modify your suggestion to align it better with this point would be the following:

> This set of characteristics can be intentionally small and can differ from the person's physical-world characteristics or [=identifiers=]. People frequently present different identities to different contexts, and also frequently share an identity among several contexts.

Does that align with where you wanted to take this @jyasskin

jyasskin: I think you've just added "or [=identifiers=]" to my suggestion (I think in order to pull in its mention of "browser fingerprints"), and that sounds good to me.

kdenhartog: Cool, updating it now and then this one should be good to go.

kdenhartog:
Updated the text with @jyasskin suggestions and resolved conflicts.

After re-reading this whole section and some of the changes made to the "Data Minimization" section as well as the new "Support Choosing Which Information to Present" section I do feel like the original intent here has basically word smithed out of this particular change, but has made it into other sections of this document. For this reason I'd be +0 to whether or not this actually gets merged at this point. Feel free to just close it if there's not consensus on the agenda when this will be discussed ned.

index.html
```diff
@@ -922,9 +922,11 @@
 
 A [=person=]'s <dfn>identity</dfn> is the set of characteristics that define
 them. Their identity *in a [=context=]* is the set of characteristics they
-present in that context. People frequently present different identities to
-different contexts, and also frequently share an identity among several
-contexts.
+present to that context, or that the context can infer.  This set of
```
darobin, line 925:
```suggestion
present to that [=context=], or that they can reasonably expect observing [=parties=] to infer from their behavior in that [=context=].  This set of
```

darobin: I like what you're saying, but:
* I don't think that contexts can infer anything (a party observing you in the context can).
* I don't want to say that any inference is potentially your identity. Maybe being gay is statistically related to, I don't know, the enjoyment of cheese.<sup>1</sup> If you're not out when you eat at my restaurant, I shouldn't consider that your orientation is part of your identity in that context even if it's pretty clear from your eating history.

<sup>1.</sup> For a time I worked on a small research team in which I was the only nongay person and the only one who didn't hate cheese. Result for small `n`, though.

darobin (changes requested):

index.html
kdenhartog: I get your point, but I think this goes against the way websites are actually abusing identity on the web today which is why I aimed for the direction I did, but we've not really landed anywhere close to it. There's numerous sites which infer attributes based on behavioral usage or specific device information which aren't reasonably expected by the user. Take for example, [this story](https://www.forbes.com/sites/kashmirhill/2012/02/16/how-target-figured-out-a-teen-girl-was-pregnant-before-her-father-did/?sh=4df7391a6668) about Target. Reasonably, no one expected Target to be able to infer this information based on purchasing behavior. And this is data that the customer could expect Target to know and collect. If they had this wouldn't be a news story. Similarly people wouldn't be constantly talking about how their microphone is listening to them when in reality that's just how good of inferential data can be derived today and no one reasonably understands it. Except this is actually what happens.

This is why I was aiming for a direction of making it clear that the browser should be acting in the users best interest in creating a default profile to remove the ability to infer as much as possible. It's become clear that's not possible as it would break webcompat in serious ways because lots of these correctable attributes have become useful to use and that's understandable. However, looking at where I started from versus where this language has ended up puts into perspective how far apart we are from where I believe many want to go (and describe principles of) versus what we can realistically state. This is what I meant by we'd wordsmithed all the meaning out of this.

kdenhartog:
Realistically I can't say I'd want to sign my name to where this text has landed compared to what I think the original intent of the issue I raised was about. I've not spent the time and effort to really understand the purpose and scope of this document nor do I think I could based on how late I decided to contribute, so I think the best thing to do is to close this. If the regular participating members of this group do believe this provides additional clarity to what I thought was missing and attempted to describe in #163 then I'd say go ahead and re-open and merge it, but I don't personally feel that we can achieve consensus based on the conversations and edits that have been made to this portion of the text. Going to close this without merging. Thank you everyone for the awesome feedback and I greatly appreciate your work on this document!

PR #167, opened 2022-05-23
add additional characteristics used to identify users

kdenhartog:
Related to the concept in PR #166 this calls out some less common characteristics which the user doesn't explicitly provide, but which should be accounted for when attempting to produce a private identity for the user when they do not wish to share anything about themselves in particular contexts.

cc @jyasskin since this one will be related to the original issue as well

***
<a href="https://pr-preview.s3.amazonaws.com/kdenhartog/privacy-principles/pull/167.html" title="Last updated on Jun 22, 2022, 4:09 PM UTC (70fa574)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/167/baea919...kdenhartog:70fa574.html" title="Last updated on Jun 22, 2022, 4:09 PM UTC (70fa574)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -884,9 +884,10 @@ store data about that [=person=]. Examples of [=identifiers=] for a [=person=] c
 * an identification number including those mapping to a device that this [=person=] may be using,
 * their phone number,
 * their location data,
-* an online identifier such as email or IP addresses, or
+* an online identifier such as email or IP addresses,
+* browser fingerprints (based on a combination of configuration characteristics) or
 * factors specific to their physical, physiological, genetic, mental, economic,
-cultural, social, behavioral, or browser [=identity=].
+cultural, social, or behavioral [=identity=].
 
 Strings derived from [=identifiers=], for instance through hashing, are still
 [=identifiers=] so long as they may identify a [=person=].
--- a/index.html
+++ b/index.html
@@ -885,8 +885,7 @@ store data about that [=person=]. Examples of [=identifiers=] for a [=person=] c
 * their phone number,
 * their location data,
 * an online identifier such as email or IP addresses, or
-* factors specific to their physical, physiological, genetic, mental, economic, cultural, or social
-[=identity=].
+* factors specific to their physical, physiological, genetic, mental, economic, cultural, social, behavioral, or browser [=identity=].
 
 Strings derived from [=identifiers=], for instance through hashing, are still
 [=identifiers=] so long as they may identify a [=person=].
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -885,8 +885,8 @@
 * their phone number,
 * their location data,
 * an online identifier such as email or IP addresses, or
-* factors specific to their physical, physiological, genetic, mental, economic, cultural, or social
-[=identity=].
+* factors specific to their physical, physiological, genetic, mental, economic,
+cultural, social, behavioral, or browser [=identity=].
```
torgo, line 889:
```suggestion
cultural, social or behavioral [=identity=].
```

darobin: @kdenhartog by "browser" did you mean cookie-based things or more like fingerprinting? If cookie-based, it feels like it's already captured by the identifier bullet but we could also add a bullet for fingerprinting data (while also mentioning unsanctioned tracking).

kdenhartog: Yeah I was meaning more like the browsers fingerprint. So more in line with if someone were to run some subset of the https://www.amiunique.org scan and assign it an identifier is the main style of tracking I wanted to call out further.

I called this the browser's identity because it's characteristics about the browser itself, not about me, but I think "browser's fingerprint identity" may be a better term to use here to further clarify.

npdoty:
I think explicitly noting browser fingerprinting (based on a combination of configuration characteristics) is a useful addition to this section. That would probably be more readable as a separate bullet point, as it's a pretty distinct thing from social and cultural details.

kdenhartog:
> I think explicitly noting browser fingerprinting (based on a combination of configuration characteristics) is a useful addition to this section. That would probably be more readable as a separate bullet point, as it's a pretty distinct thing from social and cultural details.

This is how I had thought I had originally wrote it and I may have just accepted an edit in order to combine them. 

curious what @jyasskin @darobin (editors) and @torgo (suggested changes to this) would think of separating this out?

kdenhartog:
Ok, I've made some updates to the text to separate that out now.

index.html
```diff
@@ -884,9 +884,10 @@
 * an identification number including those mapping to a device that this [=person=] may be using,
 * their phone number,
 * their location data,
-* an online identifier such as email or IP addresses, or
-* factors specific to their physical, physiological, genetic, mental, economic, cultural, or social
-[=identity=].
+* an online identifier such as email or IP addresses,
+* browser fingerprints (based on a combination of configuration characteristics) or
```
darobin, line 888:
```suggestion
* browser fingerprints (based on a combination of configuration characteristics), or
```

Issue #165, opened 2022-05-16
Contributing to an aggregate


martinthomson:
This might be part of the answer to what #150 touches on.

In some cases, there might be some ancillary or collective privacy benefit to be gained by producing some limited amount of information, even if that information would otherwise be private.

Examples that are perhaps less controversial are how browser telemetry is collected in order for browser makers to better understand how browsers are used and thereby improve them.  Another example is the collection of (unattributed) information about sites that engage in practices that might be counter to privacy (such as fingerprinting) or security (exploitation of bugs).

The document currently talks about the value of collective governance, but does not address the potential role of collective information in the system.

There is a technological aspect to this question in that many of the options we might consider are enabled by new techniques or confounded by the lack of feasible technical options.

This is something that @patcg is discussing in the context of providing better means of measuring advertising.  Some input from this group on the topic might help, though it would naturally be better as a conversation than a proclamation.  I know that many of the people involved here are engaged in that work.

torgo:
We think #184 may have addressed this issue.

torgo:
Agreed to close due to imminent merger of #216.

Issue #121, opened 2022-02-03
Discuss aggregate statistics about site users


jyasskin:
Sites often want to know how many people are reading which articles, or other aggregate information about their audiences. This is fairly benign when it's just a total number of users over the whole site, but it can become harmful when it starts letting advertisers on the site target to just the demographic that does or doesn't visit the site.

Collecting aggregate statistics also implies that the site can distinguish different people and learn some of their characteristics, which is currently accomplished by attaching those characteristics to individual identified people. There's hope that it'll be possible to do this more-privately in the long run.

We can help motivate other groups to develop the more-private mechanisms if we identify what kinds of aggregate statistics are ok and what kinds aren't.

darobin:
Needs @jyasskin.

npdoty:
https://w3ctag.github.io/privacy-principles/#ancillary-uses includes text now on aggregation, why it might be more acceptable, and notes about where ongoing work is happening

PR #170, opened 2022-06-14
[overtaken] discussion text for ancillary uses / aggregation e.g. #150

npdoty:
This is text just for discussion. It re-writes this section rather than making a small diff, just to make comparison easier; it might need to be revised or merged with existing text or made into separate sections if we want to pursue this approach.

make principles about minimization and about asking users
citations to Data Minimization finding
listing current purposes UAs share outside of particular navigations
senses of when these might be acceptable/willing/supportive
why aggregation is useful for collective purposes, but not complete

This tries to take or address the points from #121 and #165 as well.

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/pull/170.html" title="Last updated on Sep 28, 2022, 10:10 PM UTC (d195761)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/170/1fdfb27...d195761.html" title="Last updated on Sep 28, 2022, 10:10 PM UTC (d195761)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1037,10 +1037,6 @@ of an event that occurs nearly-simultaneously on both sites.
 ## Data minimization {#data}
 
 <aside class="issue">
-  This section has not been edited for narrative.
-</aside>
-
-<aside class="issue">
   There is not currently consensus among the authors and contributors to this document about the below
   principles. Please see <a href="https://github.com/w3ctag/privacy-principles/issues/150">this
   issue</a>
@@ -1095,7 +1091,23 @@ such a use) and some people would identify those uses as supporting their own go
 Aggregated or <a href="#dfn-de-identified">de-identified</a> data is often more likely to be
 acceptable in these senses and often still contributes to the collective benefit while minimizing
 privacy threats to a particular individual (see <a href="#principle-collective-privacy">collective
-privacy</a>). But some users might also object, in particular instances or in general.
+privacy</a>). But some people might also object, in particular instances or in general.
+
+<aside class="example">
+  Privacy-preserving measurement techniques may be used for aggregate calculations while minimizing
+  the parties that have access to personal data about many individual people. Encryption and
+  privacy-preserving proxies may be used to minimize the parties that have access to personal data or
+  to hide the contents of personal data used for telemetry or other kinds of measurement. But even
+  with those protections, some people may prefer not to participate in some kinds of measurement.
+
+  Ongoing work on privacy-preserving technologies in the <abbr title="Internet Engineering Task
+  Force">IETF</abbr> <a href="https://datatracker.ietf.org/wg/ppm/about/"><abbr
+  title="privacy-preserving measurement">ppm</abbr></a>, <abbr title="Internet Research Task
+  Force">IRTF</abbr> <a href="https://datatracker.ietf.org/rg/pearg/about/"><abbr title="Privacy
+  Enhancements and Assessments Research Group">pearg</abbr></a> and W3C <a
+  href="https://patcg.github.io/"><abbr title="Private Advertising Technology Community
+  Group">PATCG</abbr></a> groups may be relevant. 
+</aside>
 
 <span class="practicelab">Sites and user agents should seek to understand and respect people's goals
 and preferences about use of data about them.</span>
--- a/index.html
+++ b/index.html
@@ -1047,47 +1047,55 @@ of an event that occurs nearly-simultaneously on both sites.
   for more details.
 </aside>
 
-<div class="practice">
+<span class="practicelab">Sites, user agents and other parties should minimize the amount of data
+about people that is shared with or transferred between parties on the Web.</span>
 
-<p><span class="practicelab" id="user-agent-data-sharing">
-User agents should only share and present information about the user that
-the user can reasonably anticipate being shared, and which directly relate to the users overt, immediate goals.
-User agents and sites should be conservative when deciding if information is related to a users goals; the
-longer the chain of reasoning used to justify data collection, the less likely for that data
-collection to be ethical and principled.
-</span></p>
+Data minimization limits the risks of data being disclosed or misused and also makes it possible to
+more meaningfully provide people with understandable decisions about data about them.
 
-</div>
+<span class="practicelab">APIs should be designed to minimize the amount of data that is requested
+and provide granularity and user controls over personal data that is communicated to sites.</span>
 
-Examples:
+<aside class="note">
+This principle is further detailed in the TAG finding on <a
+href="https://www.w3.org/2001/tag/doc/APIMinimization-20100605.html">Data Minimization in Web
+APIs</a>.
+</aside>
 
-* Sharing geo-location data with a map website after a user has clicked a "share my location" button has
-  a small number of "hops" between intent and data.
-* Sharing information about the user's selected DNS resolver with a site, so that the site authors can debug
-  possible issues around tail/uncommon resolvers, has a large number of "hops" between user intent (to do
-  a thing on the website) and collection motivation (users like our site, so users will want to share information
-  with us to help us improve the site).
+In maintaining duties of <a href="#dfn-duty-of-protection">protection</a>, <a
+href="#dfn-duty-of-discretion">discretion</a> and <a href="#dfn-duty-of-loyalty">loyalty</a>, user
+agents should be cautious about sharing data not necessary to satisfy a user's immediate goals.
 
-<div class="practice">
+Data is often shared by user agents even outside the need to load and display a page for uses
+including: 
 
-<p><span class="practicelab" id="sensitive-vs-non-sensitive-data">
-User agents should generally not attempt to distinguish between sensitive and nonsensitive personal data; all
-data and personal information is likely to be sensitive to some users, often in ways unanticipated by
-browser vendors and site operators.
-</span></p>
+* usage reporting of sites or browser features; 
+* debugging site issues; 
+* measuring performance; 
+* detecting security problems or attacks; 
+* software updates; 
+* prefetching content. 
 
-</div>
+In some cases, mechanisms for sharing data are provided in order to minimize some other data
+collection mechanism identified as more intrusive or costly.
 
-<div class="practice">
+For some uses, some people would reasonably anticipate sharing such data and some people would be
+willing to participate (for example, might have explicitly decided to contribute if they had the
+time and understanding to consider details about such a use) and some people would identify those
+uses as supporting their own goals. 
 
-<p><span class="practicelab" id="minimal-data-sharing">
-User agents should always aim to share the minimal amount of information with a site
-that is needed for the user to achieve a user's goals. Data collected today can often be used in
-unanticipated ways in the future to enact privacy harms, even data that seems unlikely to
-be identifying, sensitive or otherwise privacy-harming.
-</span></p>
+Aggregated or <a href="#dfn-de-identified">de-identified</a> data is often more likely to be
+acceptable in these senses and often still contributes to the collective benefit while minimizing
+privacy threats to a particular individual (see <a href="#principle-collective-privacy">collective
+privacy</a>). But some users might also object, in particular instances or in general. In order to
+distinguish these cases, asking people about their goals and preferences is best.
 
-</div>
+<span class="practicelab">Sites and user agents should directly ask people about their goals and
+their preferences about use of data about them.</span>
+
+Because personal data may be sensitive in unexpected ways, or have risks of future uses that could
+be unexpected or harmful, minimization as a principle applies to personal data that is not currently
+known to be identifying, sensitive or otherwise potentially harmful.
 
 ## Sensitive Information {#hl-sensitive-information}
 
--- a/index.html
+++ b/index.html
@@ -1034,16 +1034,16 @@ collecting the same piece of identifying information on both sites, or by correl
 of an event that occurs nearly-simultaneously on both sites.
 
 
-## Personal Data {#data}
+## Data minimization {#data}
 
 <aside class="issue">
   This section has not been edited for narrative.
 </aside>
 
 <aside class="issue">
-  There is not currently consensus among the authors and contributors to this
-  document about the below principles.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1047,47 +1047,55 @@
...
-</span></p>
+Aggregated or <a href="#dfn-de-identified">de-identified</a> data is often more likely to be
+acceptable in these senses and often still contributes to the collective benefit while minimizing
+privacy threats to a particular individual (see <a href="#principle-collective-privacy">collective
+privacy</a>). But some users might also object, in particular instances or in general. In order to
+distinguish these cases, asking people about their goals and preferences is best.
 
-</div>
+<span class="practicelab">Sites and user agents should directly ask people about their goals and
+their preferences about use of data about them.</span>
```
spanicker, line 1094: Could this principle be targeted to UAs rather than also websites?
sites asking for consent can perpetuate the cookie banner problem

index.html
```diff
@@ -1047,47 +1047,55 @@
...
+uses as supporting their own goals. 
 
-<p><span class="practicelab" id="minimal-data-sharing">
-User agents should always aim to share the minimal amount of information with a site
-that is needed for the user to achieve a user's goals. Data collected today can often be used in
-unanticipated ways in the future to enact privacy harms, even data that seems unlikely to
-be identifying, sensitive or otherwise privacy-harming.
-</span></p>
+Aggregated or <a href="#dfn-de-identified">de-identified</a> data is often more likely to be
+acceptable in these senses and often still contributes to the collective benefit while minimizing
```
pes10k, line 1088: Acceptable to who?  As mentioned above, some users may approve of it, and others don't.  If users have explicitly agreed to something after being directly asked, then it seems like there is no privacy concern at all in most cases.

Whats missing from this text, and why i do no agree with the changes its proposing, is that this PR's text is silent on the tricky questions.  Specifically, what user data UA's should provide to websites before and or without asking explicit consent.

index.html
```diff
@@ -1047,47 +1047,55 @@
...
-* Sharing information about the user's selected DNS resolver with a site, so that the site authors can debug
-  possible issues around tail/uncommon resolvers, has a large number of "hops" between user intent (to do
-  a thing on the website) and collection motivation (users like our site, so users will want to share information
-  with us to help us improve the site).
+In maintaining duties of <a href="#dfn-duty-of-protection">protection</a>, <a
+href="#dfn-duty-of-discretion">discretion</a> and <a href="#dfn-duty-of-loyalty">loyalty</a>, user
+agents should be cautious about sharing data not necessary to satisfy a user's immediate goals.
 
-<div class="practice">
+Data is often shared by user agents even outside the need to load and display a page for uses
```
pes10k, line 1069: This is true, but similar to the following text, this change doesn't say anything about what _should_ be true, or what browsers should be doing.  Its replacing a normative statement / principal, with empirical observation.  If the aim is list kinds of data browsers collect thats not needed to load and display a page, I suggest also adding:
1. data collection and monitizaiton
2. user tracking and reidentification
3. etc

index.html
```diff
@@ -1047,47 +1047,55 @@
...
-the user can reasonably anticipate being shared, and which directly relate to the users overt, immediate goals.
-User agents and sites should be conservative when deciding if information is related to a users goals; the
-longer the chain of reasoning used to justify data collection, the less likely for that data
-collection to be ethical and principled.
-</span></p>
+Data minimization limits the risks of data being disclosed or misused and also makes it possible to
+more meaningfully provide people with understandable decisions about data about them.
 
-</div>
+<span class="practicelab">APIs should be designed to minimize the amount of data that is requested
```
pes10k, line 1056: minimize the amount of data that is requested by who, and for what purpose?

pes10k:
I've left some comments inline, but in general, i don't agree with this PR as an improvement to the underlying PR.  To my read, it replacement principals / normative statements with empirical observations, and doesn't address the core conflict / question the underlying PR: what kinds of data should browsers provide to sites / data collectors before users are asked, or without ever asking users.

Looking forward to discussing more on our next call

npdoty:
Thanks Pete for your review and comments here. I didn't have time to respond inline or try to provide updates to the PR yet, but happy to discuss.

To add to the comments above, here's a quick summary of the key changes I heard suggested at the last task force call where this was discussed:

* in the empirical listing of data that is communicated outside of loading a particular webpage, separate out browser telemetry (communications with the browser vendor) from communications with the site or other services
* principle should be that UAs should request general preferences on sending data on behalf of users. or perhaps, UAs should seek to understand and respect user desires, preferences and goals (which could include heuristics or understanding of trust relationship that the user has with a site)
* this should probably just be a separate section on Data Minimization, rather than Personal Data
* better to have principles about API design for site access, rather than giving specific recommendations to sites

index.html
```diff
@@ -1034,60 +1034,95 @@
...
+href="#dfn-ancillary-uses">ancillary use</a> and also not burden the user with additional [=privacy
+labor=]. To that end, user agents may employ user research, solicitation of general preferences, and
+heuristics about sensitivity of data or trust in a particular context. To facilitate site
+understanding of user preferences, user agents can provide browser-configurable signals to directly
+communicate common user preferences.
+
+<aside class="example">
+  Sites and browsers wish to collect telemetry data to determine how frequently features are used or
+  to debug breakages, but the user agent does not want to burden the user with frequent consent
+  requests. A browser could use a first-run dialog to ask the user whether they generally support
```
spanicker, line 1113: A first-run dialog for the browser installation right? Or is this for first visit on a given website?

npdoty: Ah, yes, I meant "first-run" to imply first run of a browser or piece of local software. But if that's not clear, we should wordsmith it.

And clearly this is just an example. Maybe user agents should check in with their users occasionally, rather than asking a bunch of questions on first-run/installation. We should encourage experimentation.

index.html
```diff
@@ -1034,60 +1034,107 @@
...
+href="https://www.w3.org/2001/tag/doc/APIMinimization-20100605.html">Data Minimization in Web
+APIs</a>.
+</aside>
 
-</div>
+In maintaining duties of <a href="#dfn-duty-of-protection">protection</a>, <a
+href="#dfn-duty-of-discretion">discretion</a> and <a href="#dfn-duty-of-loyalty">loyalty</a>, user
+agents should be conservative when sharing data not necessary to satisfy a user's immediate goals.
+
+On the Web, data is often shared even outside the need to load and display a page for <dfn>ancillary
```
pes10k, line 1065: Small edit: I'd like to avoid passive voice here, since it obscures who is sharing what data with whom.  How about:

> Some Web browsers share information about their users experiences, environments and computing performance, even when that information is not needed to load and display a page. Browsers may share these kinds of data with websites for a variety of <dfn>ancillary uses</dfn>, including:

index.html
```diff
@@ -1034,60 +1034,107 @@
...
+<aside class="note">
+This principle was further explored in an earlier TAG draft on <a
+href="https://www.w3.org/2001/tag/doc/APIMinimization-20100605.html">Data Minimization in Web
+APIs</a>.
+</aside>
 
-</div>
+In maintaining duties of <a href="#dfn-duty-of-protection">protection</a>, <a
+href="#dfn-duty-of-discretion">discretion</a> and <a href="#dfn-duty-of-loyalty">loyalty</a>, user
+agents should be conservative when sharing data not necessary to satisfy a user's immediate goals.
```
pes10k, line 1063: I would like to reword this, since it seems to suggest that user agents _should_ share data that doesn't benefit users, and this (to me) still reads like it is worried mostly about site goals, and not the goals of the person using the browser.

In general, if we all agree that browsers should not share this information with sites if users do not want their browser to do so, i suggest we reword it around user intent.  WDYT of this:

```
Browsers must send some amount of information to websites in order for those sites to execute
correctly, and for users to be able to accomplish their goals in interacting with those sites.  Some
users may wish to share <em>additional data</em> with websites though, data not directly related
to the users goals in interacting with the site, but which might help the site accomplish other goals.
This non-necessary data is <dfn>ancillary data</dfn> and the purposes sites wish to use it for are
<dfn>ancillary uses</dfn>.

Browsers should only send ancillary data to sites when the browser is confident the user
wishes to send such data. Browsers should be careful, cautious and conservative when making
such decisions.

Examples of the kind of ancillary data users might wish to send to sites include information
about the users experience loading the site, the kind of browser or computing environment
the site is being executed in, or the performance of the site as the user interacts with it.

Examples of the kinds of ancillary uses sites might wish to collect such data for include:

(existing list)
```

I hope the above text accomplishes the following:
1. demonstrates there are categories of non-necessary data that sites might want to collect for benign and non-nefarious purposes
2. center the user in the decision making process, and emphasize that this data should be sent because the user wants to send it, not (only) bc the site wants it
3. address @torgo 's concern about not calling for any specific mechanism the browser should use to make sure the browser is carrying out the user's goals (e.g., avoiding overloaded terms like "asking for consent" that might suggest certain UI / UX)

index.html
```diff
@@ -1034,60 +1034,107 @@
...
+  * detecting security problems or attacks; 
+* advertising and social media sharing, including: 
+  * measurement, targeting, ad auctions, personalization, fraud prevention; 
+* performance, including caching and prefetching content; 
+* software updates.
+
+In some cases, mechanisms for sharing data are provided in order to minimize some other data
+collection mechanism identified as more intrusive or costly.
+
+For some uses, some people would reasonably anticipate sharing such data or identify that use as a
```
pes10k, line 1086: This text reads odd to me; I understand the goal of the text to be "different things are acceptable and desireable to different people", but I think something else needs to be added to make that into an actionable principal.

WDYT of prefacing with something like "Different users will want to share different kinds and amounts of data with websites. (current paragraph here). Browsers should only share data with websites that is acceptable and desirable to each particular browser user."?

index.html
```diff
@@ -1034,60 +1034,107 @@
...
+  * performance measurements, 
+  * measuring feature usage, 
+  * debugging site, browser and networking issues, and, 
+  * detecting security problems or attacks; 
+* site telemetry, including: 
+  * performance measurements, 
+  * analytics about usage and visitors, 
+  * debugging site issues, and, 
+  * detecting security problems or attacks; 
+* advertising and social media sharing, including:
```
spanicker, line 1078: advertising use-case that targets users is quite different from the list of straight-up telemetry use-cases, that don't involve any personal data or device identifiers.
Might make sense to split these?

index.html
spanicker: Agree that the text can be clarified. 
I disagree with the suggestion though. Sensible defaults should be possible and UAs should not have to offload this choice entirely to each particular user.

index.html
```diff
@@ -1034,60 +1034,107 @@
...
-longer the chain of reasoning used to justify data collection, the less likely for that data
-collection to be ethical and principled.
-</span></p>
+Data minimization limits the risks of data being disclosed or misused and also makes it possible to
+more meaningfully provide people with understandable decisions about data about them.
 
-</div>
-
-Examples:
+<span class="practicelab">Web APIs should be designed to minimize the amount of data that is requested
```
jyasskin, line 1052: How much can we avoid passive voice here?

```suggestion
<span class="practicelab">Web APIs should be designed to minimize the amount of data that sites need to request to carry out their users' goals
```

would be a start ... "granularity" might be redundant since sites can only request a minimal amount of information if the grains of information are small.

Then ... do user controls belong in this principle? I could imagine saying that we should try to design APIs to maximize the amount of user control that UAs can provide, but that's not really about minimization anymore. More about #181 maybe.

index.html
jyasskin: We should draw from https://www.yalelawjournal.org/feature/a-relational-theory-of-data-governance in wording this. Pete's suggested text is very individualistic, saying not to share unless it's worth it to an individual user to specifically opt into the sharing. But Professor Viljoen points out that we have collective interests in aggregating data, which should be managed socially rather than just individually. An open standards body like the W3C isn't exactly democratic, but it's closer than either just browsers or just individual users.

yoavweiss:
I think it'd be worthwhile to also factor in the privacy cost that any web exposed user preference has into this text: web-exposing a user's opt-in/out from ancillary uses would also provide a persistent fingerprinting vector that can be used for their re-identification, and that would need to be mitigated in some way.

It'd be similarly worthwhile to clarify the user benefits from being able to opt-in/out of ancillary data uses.

index.html
npdoty: As we've discussed a bit on previous calls, I don't think we want to define a category of data as ancillary data. Data can be used for multiple purposes, and it could be the same pieces of data, or a single API could be used for functional purposes on one site and for an ancillary use on another site.

(I'm trying to take other parts of this wording for an updated PR.)

npdoty:
#183 and #184 represented different parts of this proposed PR. It's likely we'll continue to iterate on those sections.

I'm not closing this PR yet in case there are discussions here that haven't been resolved or moved elsewhere.

npdoty:
@yoavweiss 
> I think it'd be worthwhile to also factor in the privacy cost that any web exposed user preference has into this text: web-exposing a user's opt-in/out from ancillary uses would also provide a persistent fingerprinting vector that can be used for their re-identification, and that would need to be mitigated in some way.

I've noted fingerprinting surface that you and @pes10k noted, that might be available in data commonly used for telemetry or in any revelation of user preferences about telemetry. (That text is in #184, now merged, but still iterating.)

There is an open question (from @torgo) about what guidance we should give about disabling features or otherwise not revealing that a mode not sending telemetry or analytics data is engaged. Related platform design principles include:
* https://w3ctag.github.io/design-principles/#feature-detect
* https://w3ctag.github.io/design-principles/#do-not-expose-use-of-private-browsing-mode

torgo:
Closed as agreed on today's call.

yoavweiss:
> Closed as agreed on today's call.

Any link to the minutes?

jyasskin:
The Nov 2 minutes are https://github.com/w3ctag/privacy-principles/blob/main/meetings/2022-11-02-minutes.md, but I think the actual explanation is that this was closed because its content was merged as part of #183 and #184. See https://w3ctag.github.io/privacy-principles/#data-minimization for the resulting text, and I think further discussion belongs in #150.

PR #181, opened 2022-08-31
Add a principle to support deception (fixes #174)

### LINKED ISSUE #174: UAs should support deception principle
It would be good if browsers helped people lie, for instance enter arbitrary geolocation or use a throwaway ID. @rhiaro had mentioned this as relevant: https://dr.amy.gy/chapter3#computationally-mediated-pro-social-deception

darobin:
***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/pull/181.html#support-deception" title="Last updated on Oct 26, 2022, 4:20 PM UTC (b237e23)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/181/76b45e0...b237e23.html" title="Last updated on Oct 26, 2022, 4:20 PM UTC (b237e23)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1546,7 +1546,9 @@ to respond with information that may differ from the truth can be essential in p
 component of fingerprinting mitigation, and can be required for auditing purposes such that an auditor may
 pretend to be someone else in order for instance to detect discrimination.
 
-In some cases, sites do not abide by data minimisation principles and request more information than they realistically require. This principle supports [=people=] in defending under such cases.
+In some cases, sites do not abide by data minimisation principles and request more information than they realistically
+require. This principle supports [=people=] in defending under such cases.
+
 It is worth noting that this principle deliberately empowers [=people=] to provide deceptive information. In the
 general case, the ability for people to present as they wish is more important than the ability of computers to
 prevent them from doing so. Sites should include deception in their threat modelling.
--- a/index.html
+++ b/index.html
@@ -1515,12 +1515,13 @@ trick the [=person=] into opting back into the [=processing=].
 
 <aside class="issue">Some services have the user pay for their use in data. These services aren't necessarily retaliating by denying their services to users who refuse to pay with data, but the details are more complex than we've had time to write.</aside>
 
-## Support Deception {#support-deception}
+## Support Choosing Which Information to Present {#support-choosing-info}
 
 <div class="practice">
-  <span class="practicelab" id="principle-support-deception">
-    [=User agents=] should support [=people=] in providing false or incorrect information to [=parties=]
-    that request it.
+  <span class="practicelab" id="principle-support-choosing-info">
+    [=User agents=] should support [=people=] in choosing which information they provide to [=parties=] that
+    request it, up to and including allowing for the provision of information that may not be factually
+    correct.
   </span>
 </div>
 
@@ -1528,13 +1529,30 @@ trick the [=person=] into opting back into the [=processing=].
 design their products in ways that make it a lot easier for [=people=] to disclose information than not, whereas
 [=people=] typically have to manually wade through options, repeated prompts, and deceptive patterns. In many
 cases, the absence of data — when a [=person=] refuses to provide some information — can also be identifying
-or revealing. [=User agents=] should make it simple for [=people=] to evade such tactics and to live in
-obscurity ([[?Lost-In-Crowd]], [[?Obscurity-By-Design]]) by rendering the provision of incorrect information as easy as
-possible ([[?Obfuscation]]).
+or revealing. Additionally, APIs can be defined or implemented in rigid ways that can prevent people from
+accessing useful functionality (eg. if my geolocation is forcefully set to match my GPS, a restaurant-finding
+site might restrict me to my current location when I want to look for restaurants in a city I will be visiting
+this weekend).
 
-Examples of ways in which a [=user agent=] may support its [=user=]'s obscurity include:
-* Generating domain-specific email addresses so that [=people=] can log into the site without becoming recognisable
-  across contexts.
+[=User agents=] should make it simple for [=people=] to present the identity they wish to and to provide information
+about themselves or their devices in ways that they control. This empowers [=people=] to live in obscurity
+([[?Lost-In-Crowd]], [[?Obscurity-By-Design]]) by rendering the provision of information according to their own
+preferences as easy as possible ([[?Obfuscation]]).
+
+APIs should be designed such that they do not expect to provide the truth about a [=person=] as provided for
+instance by a sensor or recorded data. Instead, the API could indicate a [=person=]'s preference, a [=person=]'s
+chosen identity, a [=person=]'s query or interest, or a [=person=]'s selected communication style. Directing APIs
+to respond with information that may differ from the truth can be essential in protecting [=people=], is a key
+component of fingerprinting mitigation, and can be required for auditing purposes such that an auditor may
+pretend to be someone else in order for instance to detect discrimination.
+
+It is worth noting that this principle deliberately empowers [=people=] to provide deceptive information. In the
+general case, the ability for people to present as they wish is more important than the ability of computers to
+prevent them from doing so. Sites should include deception in their threat modelling.
+
+Examples of ways in which a [=user agent=] may support this principle include:
+* Generating domain-specific email addresses or other directed identifiers so that [=people=] can log into the
+  site without becoming recognisable across contexts.
 * Offering the option to produce fake geolocation and accelerometry data, with parameters specified by the [=user=].
 * Making it easy to upload a picture while pretending to have it taken by the camera.
 * Faking the acceptance of notification prompts (or pretending that notifications have already been accepted).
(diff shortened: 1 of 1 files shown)


darobin:
Minutes: https://cryptpad.w3ctag.org/code/#/2/code/edit/goYdOL8HNopHSharasC2whlQ/

darobin:
From chat:

* Nick Doty: I think a good goal would be for very few APIs to be defined as asserting a particular truth about the user and instead APIs could indicate a user’s preference, a user’s chosen identity, a user’s query or interest, a user’s selected communication style, and in the exceedingly rare cases where an API must be designed to assert a truth, I would support the ability of users to lie (even if software may not always be designed to help them lie) and also the guidance to sites that users may lie. and actually, maybe I think those cases should never exist. if some other party wants to assert something about the user, they can do so, and the recipient can decide whether or not to trust it
* Peter: (I like nick’s comment above, so some of this is probably agreement now, i just wanted to follow up with what i was going to say on the call). Quick, incomplete-list of pro-deception arguments (and caveat all this with me acknowledging that “deception vs my preferred identity” is probably 90% the same if scratched enough”)
  1. sites can (and in some cases, already do) reject “directed identifiers” if they can detect they’re being used.  In these cases, browsers should help the user lie about whether the user is using a directed-identifier
  1. lying about API responses is an important anti-fingerprinting defense (even in cases where the spec says the browser should/must reply with the actual hardware capability / underlying value / etc)
  1. Auditing (for legal and anti-discrimination uses, among others) often requests the auditor / user to try to sincerely convince the site they are someone the user isn’t.  In these cases, the goal isn’t just “presenting a different identity”, its intentional deception
* Don: Another one -- The ability to generate believable fake answers to norms-violating questions tends to reduce the quality of data obtained by asking those questions, so could reduce the ROI of investing in asking them. (imho the role of privacy tech is only partly individual user protection -- it's more important in the long run as a way to incentivize future investors and decision-makers to make good choices)
* Peter: Nick's point about “API’s should avoid being defined in such a way that they require ‘true responses’” is a very good one though, and would be a great addition to the Design Principals and S&P questionnaire docs

index.html
```diff
@@ -1498,6 +1515,50 @@
...
+([[?Lost-In-Crowd]], [[?Obscurity-By-Design]]) by rendering the provision of information according to their own
+preferences as easy as possible ([[?Obfuscation]]).
+
+APIs should be designed such that they do not expect to provide the truth about a [=person=] as provided for
+instance by a sensor or recorded data. Instead, the API could indicate a [=person=]'s preference, a [=person=]'s
+chosen identity, a [=person=]'s query or interest, or a [=person=]'s selected communication style. Directing APIs
+to respond with information that may differ from the truth can be essential in protecting [=people=], is a key
+component of fingerprinting mitigation, and can be required for auditing purposes such that an auditor may
+pretend to be someone else in order for instance to detect discrimination.
+
```
darobin, line 1548:
```suggestion

In some cases, sites do not abide by data minimisation principles and request more information than they realistically require. This principle supports [=people=] in defending under such cases.
```

darobin: @sandandsnow Like this?

index.html
```diff
@@ -1498,6 +1515,51 @@
...
+In some cases, sites do not abide by data minimisation principles and request more information than they realistically require. This principle supports [=people=] in defending under such cases.
```
darobin, line 1549:
```suggestion
In some cases, sites do not abide by data minimisation principles and request more information than they realistically
require. This principle supports [=people=] in defending under such cases.

```

index.html
```diff
@@ -1498,6 +1515,50 @@
...
+pretend to be someone else in order for instance to detect discrimination.
+
+It is worth noting that this principle deliberately empowers [=people=] to provide deceptive information. In the
+general case, the ability for people to present as they wish is more important than the ability of computers to
+prevent them from doing so. Sites should include deception in their threat modelling.
+
+Examples of ways in which a [=user agent=] may support this principle include:
+* Generating domain-specific email addresses or other directed identifiers so that [=people=] can log into the
+  site without becoming recognisable across contexts.
+* Offering the option to produce fake geolocation and accelerometry data, with parameters specified by the [=user=].
```
rhiaro, line 1559:
```suggestion
* Offering the option to generate geolocation and accelerometry data with parameters specified by the [=user=].
```

index.html
```diff
@@ -1498,6 +1515,53 @@
...
+accessing useful functionality (eg. if my geolocation is forcefully set to match my GPS, a restaurant-finding
+site might restrict me to my current location when I want to look for restaurants in a city I will be visiting
+this weekend).
+
+[=User agents=] should make it simple for [=people=] to present the identity they wish to and to provide information
+about themselves or their devices in ways that they control. This empowers [=people=] to live in obscurity
+([[?Lost-In-Crowd]], [[?Obscurity-By-Design]]) by rendering the provision of information according to their own
+preferences as easy as possible ([[?Obfuscation]]).
+
+APIs should be designed such that they do not expect to provide the truth about a [=person=] as provided for
```
npdoty, line 1542: I think this first sentence should be a principle, and maybe even the only principle in this section; it provides direct advice about how things should be done in the design of the web platform.

Maybe:
_APIs should be designed such that data returned through an API does not assert a fact about the user or their environment._

And then that can be followed by the for instance and instead clauses, which I think are all great examples.

npdoty: Maybe "assert a fact or make a promise on the user's behalf about the user or their environment"

index.html
```diff
@@ -1498,6 +1515,53 @@
...
+instance by a sensor or recorded data. Instead, the API could indicate a [=person=]'s preference, a [=person=]'s
+chosen identity, a [=person=]'s query or interest, or a [=person=]'s selected communication style. Directing APIs
+to respond with information that may differ from the truth can be essential in protecting [=people=], is a key
```
npdoty, line 1545: Maybe:
"In any rare instances when an API must be defined as returning true current values, users may still configure their agents to respond with other information, for reasons including auditing or mitigating browser fingerprinting or other forms of data collection."

I believe our goal is that this should be an exceedingly rare situation where assertions of true facts is necessary to the design. I also disagree that it's a "key component" of fingerprinting mitigation. There has been some use of returning false or randomized data for fingerprinting mitigation, but I think by far the more common mitigations have been to reduce the data returned or the variation in that data. I don't know that most browser fingerprinting mitigations include returning false or randomized data (though that could change).

index.html
```diff
@@ -1498,6 +1515,53 @@
...
+component of fingerprinting mitigation, and can be required for auditing purposes such that an auditor may
+pretend to be someone else in order for instance to detect discrimination.
+
+In some cases, sites do not abide by data minimisation principles and request more information than they realistically
+require. This principle supports [=people=] in defending under such cases.
+
+It is worth noting that this principle deliberately empowers [=people=] to provide deceptive information. In the
```
npdoty, line 1552: Maybe:

> Sites should include deception in their threat modeling and not assume that Web platform APIs provide any guarantees of consistency, currency or correctness about the user. Users often have control of the devices and software they use to interact with web sites. In response to site requests, people may provide information that may be deceptive for a variety of reasons, including both malice and self-protection.

index.html
```diff
@@ -1498,6 +1515,53 @@
...
+prevent them from doing so. Sites should include deception in their threat modelling.
+
+Examples of ways in which a [=user agent=] may support this principle include:
+* Generating domain-specific email addresses or other directed identifiers so that [=people=] can log into the
+  site without becoming recognisable across contexts.
+* Offering the option to produce fake geolocation and accelerometry data, with parameters specified by the [=user=].
+* Making it easy to upload a picture while pretending to have it taken by the camera.
+* Faking the acceptance of notification prompts (or pretending that notifications have already been accepted).
+* Filtering outbound network traffic so as to replace specific strings (TCF, hashed email, navigation identifiers).
+* Randomly swapping third-party cookies with the browsers of other [=people=].
```
npdoty, line 1563: I'm not sure this is a helpful example for the reader. 

Third-party cookies may not always be widely supported. Swapping of cookies could be an obfuscation method, but could also lead to security vulnerabilities.

index.html
```diff
@@ -1498,6 +1515,53 @@
...
+
+It is worth noting that this principle deliberately empowers [=people=] to provide deceptive information. In the
+general case, the ability for people to present as they wish is more important than the ability of computers to
+prevent them from doing so. Sites should include deception in their threat modelling.
+
+Examples of ways in which a [=user agent=] may support this principle include:
+* Generating domain-specific email addresses or other directed identifiers so that [=people=] can log into the
+  site without becoming recognisable across contexts.
+* Offering the option to produce fake geolocation and accelerometry data, with parameters specified by the [=user=].
+* Making it easy to upload a picture while pretending to have it taken by the camera.
```
npdoty, line 1560: Maybe:
> Uploading a stored video stream in response to a camera prompt.

index.html
```diff
@@ -1498,6 +1515,53 @@
...
+* Faking the acceptance of notification prompts (or pretending that notifications have already been accepted).
```
npdoty, line 1561: Maybe:
> Automatically granting or denying permission prompts based on user configuration.

npdoty (commented):
Per my comments on previous calls, I think our principle should be to design APIs such that no assertions or promises are made by a user agent on the users behalf, and then questions about deception become a rare corner case. I've suggested a few changes here that can more consistently follow that direction, including saying that those cases are rare, and modifying examples to be about the user's choice rather than about deception.

PR #182, opened 2022-09-21
Try to explain how new APIs can or can't be justified based on old ones.

jyasskin:
***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/182.html#information" title="Last updated on Dec 14, 2022, 5:20 PM UTC (4734397)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/182/2380f22...jyasskin:4734397.html" title="Last updated on Dec 14, 2022, 5:20 PM UTC (4734397)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1317,7 +1317,7 @@ information into three categories:
     * The URL a person is visiting must be sent to a server in order to navigate to that
        URL, and known private-information-retrieval methods are too expensive to avoid that
        exposure.
-    * The [[[largest-contentful-paint]]] distribution for a website is about a website
+    * The distribution of [[[largest-contentful-paint]]] timings for a website is about a website
        rather than about the people browsing it, even if the data that informs that
        measure can also reveal information about people.
 
@@ -1357,12 +1357,12 @@ found it</a>.
 </div>
 
 Acceptable information exposure is always qualified by the (possibly empty) set of
-user-controlled settings or permissions that guard access to it. For example, the URLs of
-resources, the timing of link clicks, and the referrer chain within a single origin are
-not guarded by anything; the scroll position is guarded by the setting to turn off
+user-controlled settings or permissions that <dfn data-lt="access guard">guard</dfn> access to it.
+For example, the URLs of resources, the timing of link clicks, and the referrer chain within a
+single origin are not guarded by anything; the scroll position is guarded by the setting to turn off
 javascript; and access to the camera or geolocation are guarded by permission prompts.
 
-Information that would be acceptable to expose under one set of guards might be
+Information that would be acceptable to expose under one set of [=access guards=] might be
 unacceptable under another set, so when an API designer intends to explain that their new
 API is acceptable because an existing acceptable API already exposes the same information,
 they must be careful to ensure that their new API is only available under a set of guards
@@ -1374,7 +1374,7 @@ scratch, without relying on the existing API.
 <div class="practice">
   <span class="practicelab">If existing APIs provide access to some information, but
   we have a plan to change those APIs to prevent that access, new APIs must not be added
-  that provide that same information without extra guards that make the information
+  that provide that same information without extra [=access guards=] that make the information
   acceptable.</span>
 </div>
 
@@ -1382,7 +1382,8 @@ scratch, without relying on the existing API.
 
 <div class="practice">
   <span class="practicelab">New APIs that provide access to undesirable information should
-  not make that information easier to access.</span>
+  not make that information easier to access, unless they add [=access guards=] that make the
+  information acceptable.</span>
 </div>
 
 
@@ -1395,8 +1396,9 @@ scratch, without relying on the existing API.
    be clear how the more-private browser can also prevent the new API from exposing that
    information without breaking additional sites or user experiences.
 
-1. A new API should be harder to use to discover undesirable information than the existing
-   APIs, for example by requiring more code, more maintenance, and/or more runtime cost.
+1. When a developer is trying to access the undesirable information, a new API should be at least as
+   difficult to use as the existing APIs. For example, it shouldn't require less code, less
+   maintenance, or less runtime cost.
 
 The third consideration can be surprising. In many other cases, we can think in terms of a
 threat model, using designs familiar from security to make information either available or
--- a/index.html
+++ b/index.html
@@ -1393,7 +1393,7 @@ scratch, without relying on the existing API.
 1. If an existing browser does block access to the undesirable information, perhaps by
    breaking some experiences on the Web that other browsers don't wish to break, it should
    be clear how the more-private browser can also prevent the new API from exposing that
-   information.
+   information without breaking additional sites or user experiences.
 
 1. A new API should be harder to use to discover undesirable information than the existing
    APIs, for example by requiring more code, more maintenance, and/or more runtime cost.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1183,6 +1183,110 @@
...
+
+1. Information that's fine to expose, for example because a person or group with
+    sufficient authority intended to expose that information or to do something that
+    necessarily exposes the information, or because it's not about people at all. For
+    example:
+
+    * The geolocation and camera APIs ask whether a person wants to expose their data.
+    * The URL a person is visiting must be sent to a server in order to navigate to that
+       URL, and known private-information-retrieval methods are too expensive to avoid that
+       exposure.
```
annevk, line 1200: What does "known private-information-retrieval methods" mean here?

It seems that at least some user agents are interested in potentially modifying the URL before sending it onward and some of that would not be expensive. (It remains to be seen whether it will be effective.)

index.html
```diff
@@ -1183,6 +1183,110 @@
...
+
+<div class="practice">
+  <span class="practicelab">New APIs can add additional ways of getting acceptable
+  information that are guarded at least as strongly as the old ways.</span>
+</div>
+
+Acceptable information exposure is always qualified by the (possibly empty) set of
+user-controlled settings or permissions that guard access to it. For example, the URLs of
+resources, the timing of link clicks, and the referrer chain within a single origin are
+not guarded by anything; the scroll position is guarded by the setting to turn off
```
annevk, line 1236: I don't think referrer is that simple? A new feature would have to at least abide by Referrer Policy (and some user agents might also allow disabling it always?).

annevk (commented):
Thanks for writing this down! I think I generally agree with the classification and it's kinda in line with how I've approached API design for the web platform.

What might be rather tricky is getting agreement on what is fine to expose and what is not.

index.html
```diff
@@ -1183,6 +1183,110 @@
...
+    to. For example, browsers are gradually removing the ability to join identities
+    between different [=partitions=].
+
+1. Information that we'd rather not expose, but that we don't have a plausible plan for
+    removing access to. For example:
+
+    * Some users are disappointed that the page they're visiting can discover which link
+        they clicked to leave that page. We can't block that information that information
+        because the page can use <a data-cite="RFC9110#status.3xx">HTTP redirects</a> to
+        learn it, and redirection is a core feature of the web.
```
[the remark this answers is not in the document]
johannhof: That seems not impossible, but quite implausible to me personally :)

index.html
```diff
@@ -1183,6 +1183,110 @@
...
+   undesirable information, it should be clear how to extend those changes to the proposed
+   API.
+
+1. If an existing browser does block access to the undesirable information, perhaps by
+   breaking some experiences on the Web that other browsers don't wish to break, it should
+   be clear how the more-private browser can also prevent the new API from exposing that
+   information.
+
+1. A new API should be harder to use to discover undesirable information than the existing
+   APIs, for example by requiring more code, more maintenance, and/or more runtime cost.
```
johannhof, line 1273: I think this is a good idea but I don't think the examples are particularly fitting, which confused me for a second. In my view we don't want to design APIs that, say, enable only dedicated attackers that will have no issue sending over 100KB of JS that spends 10 seconds to compute joint identities. We want to make attacks harder in ways that can't be scaled with computing and engineering resources, such as increasing the site usage time needed until sufficient information can be gathered from a single user, increasing user effort and user exposure or by requiring increasingly tight cooperation with other (malicious) parties.

index.html
jyasskin: I'm thinking of https://en.wikipedia.org/wiki/Private_information_retrieval, which makes it theoretically possible to retrieve a URL without telling the server which URL you're getting, but IIUC requires the server to compute over their entire space of resources in order to return one.

Modifying the URL doesn't seem to solve the problem of the server learning which resource the user wants. I guess you could modify it into pointing at a different resource, but users seem unlikely to be happy with that. 🤪

index.html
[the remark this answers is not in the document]
jyasskin: I added a note below about having to balance the harm to users from exposing a bit of information against the harm to users of not-exposing it, for example by transferring many times the amount of desired information or losing the benefits of scripting.

index.html
jyasskin: In existing UAs, are same-origin referrers actually guarded by any user setting or permission? Referrer Policy, for example, is controlled by the site, so it doesn't in itself constitute a user setting. This paragraph is trying to give readers a sense of how we have to keep existing guards in new features, and I don't think speculating about future UA behavior changes helps with that.

index.html
jyasskin: The situation in this section is that we have an existing API that lets attackers send 100KB of JS and spend 10s, and someone's proposing a new API that still exposes the information, but needs 1MB of JS or 100s. Sure it'd be nice to make the existing API more difficult too, but the other part of this section is that nobody has figured out how to do that.

Clearly I didn't explain that well ... any ideas how to do it better?

johannhof: Hmm do you think that the computational effort really makes a difference in favor of the new API here? I don't really see why it should (and to be honest I'm also struggling to come up with an example where we would encounter this kind of situation in practice).

Overall if the first two sub-points here are adhered to I don't see why an API would have to be additionally strictly more computationally expensive (and how that would even be measured).

index.html
```diff
@@ -1183,6 +1183,117 @@
...
+</div>
+
+
+1. If future web platform changes make it possible to remove other access to the
+   undesirable information, it should be clear how to extend those changes to the proposed
+   API.
+
+1. If an existing browser does block access to the undesirable information, perhaps by
+   breaking some experiences on the Web that other browsers don't wish to break, it should
+   be clear how the more-private browser can also prevent the new API from exposing that
```
pes10k, line 1276: WDYT of changing

> it should be clear how the more-private browser can also prevent the new API from exposing that information.

to 

> it should be clear how the more-private browser can also prevent the new API from exposing that information _without breaking additional sites or user experiences_.

I think some text tweak is needed here because its not usually difficult (unclear) to figure out how to prevent the new API from leaking the existing information; whats difficult is judging and minimizing the compat risks of blocking the new API

jyasskin: SGTM, done.

index.html
```diff
@@ -1183,6 +1183,117 @@
...
+### Handling information that's being removed {#unacceptable-information}
+
+<div class="practice">
+  <span class="practicelab">If existing APIs provide access to some information, but
+  we have a plan to change those APIs to prevent that access, new APIs must not be added
+  that provide that same information without extra guards that make the information
+  acceptable.</span>
+</div>
+
+### Handling information we can't completely block {#sketchy-information}
```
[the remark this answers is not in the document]
jyasskin: Good point; I've added a `<dfn>` for "access guards" and added the same text in this principle that's in the "being removed" principle, that extra guards can make the information acceptable.

index.html
jyasskin: I think you're right that it just has to be not-easier, so I've changed the text in that direction. For concrete examples, `<a ping>` clearly reduced the cost of tracking clicks, since it removed the round trips from going through a redirect. Vs using a screencast API to detect adblocking, which is more difficult than looking for deleted DOM nodes or missing server-side requests.

torgo (approved):
lgtm

PR #191, opened 2022-11-02
no-assertion

npdoty:
addresses #174 

Revisions off of #181, to emphasize that APIs should be designed so that they do not assert or promise on the user's behalf, and that sites should include deception in their threat modeling, as it's likely to happen for a variety of reasons.

Still includes, in some cases, user agents supporting the ability of a user to provide information that may not be factually correct.

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/pull/191.html#support-choosing-info" title="Last updated on Nov 3, 2022, 11:05 PM UTC (b9e5200)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/191/caf9e8b...b9e5200.html" title="Last updated on Nov 3, 2022, 11:05 PM UTC (b9e5200)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -204,6 +204,12 @@
           href: 'https://www.cambridge.org/core/books/industry-unbound/787989F90DBFC08E47546178A7AB04F7',
           publisher: 'Cambridge University Press',
         },
+        'Lost-In-Crowd': {
+          title: 'Why You Can No Longer Get Lost in the Crowd',
+          authors: ['Woodrow Hartzog', 'Evan Selinger'],
+          href: 'https://www.nytimes.com/2019/04/17/opinion/data-privacy.html',
+          publisher: 'The New York Times',
+        },
         'NIST-800-63A': {
           title: 'Digital Identity Guidelines: Enrollment and Identity Proofing Requirements',
           href: 'https://pages.nist.gov/800-63-3/sp800-63a.html',
@@ -216,6 +222,17 @@
           author: ['Robin Berjon'],
           href: 'https://open.nytimes.com/how-the-new-york-times-thinks-about-your-privacy-bc07d2171531',
           publisher: 'NYT Open',
+        },
+        'Obfuscation': {
+          title: 'Obfuscation: A User\'s Guide for Privacy and Protest',
+          authors: ['Finn Brunton', 'Helen Nissenbaum'],
+          href: 'https://www.penguinrandomhouse.com/books/657301/obfuscation-by-finn-brunton-and-helen-nissenbaum/',
+          publisher: 'Penguin Random House',
+        },
+        'Obscurity-By-Design': {
+          title: 'Obscurity by Design',
+          authors: ['Woodrow Hartzog', 'Frederic Stutzman'],
+          href: 'https://papers.ssrn.com/sol3/papers.cfm?abstract_id=2284583',
         },
         'OECD-Guidelines': {
           title: 'OECD Guidelines on the Protection of Privacy and Transborder Flows of Personal Data',
@@ -1571,6 +1588,60 @@ unrelated feature, by decreasing the quality of the service, or by trying to caj
 trick the [=person=] into opting back into the [=processing=].
 
 <aside class="issue">Some services have the user pay for their use in data. These services aren't necessarily retaliating by denying their services to users who refuse to pay with data, but the details are more complex than we've had time to write.</aside>
+
+## Support Choosing Which Information to Present {#support-choosing-info}
+
+<div class="practice">
+  <span class="practicelab" id="principle-support-choosing-info">
+    [=User agents=] should support [=people=] in choosing which information they provide to [=parties=] that
+    request it, up to and including allowing for the provision of information that may not be factually
+    correct.
+  </span>
+</div>
+
+[=Parties=] can invest time and energy into automating ways of gathering [=data=] from [=people=] and can
+design their products in ways that make it a lot easier for [=people=] to disclose information than not, whereas
+[=people=] typically have to manually wade through options, repeated prompts, and deceptive patterns. In many
+cases, the absence of data — when a [=person=] refuses to provide some information — can also be identifying
+or revealing. Additionally, APIs can be defined or implemented in rigid ways that can prevent people from
+accessing useful functionality (eg. if my geolocation is forcefully set to match my GPS, a restaurant-finding
+site might restrict me to my current location when I want to look for restaurants in a city I will be visiting
+this weekend).
+
+[=User agents=] should make it simple for [=people=] to present the identity they wish to and to provide information
+about themselves or their devices in ways that they control. This empowers [=people=] to live in obscurity
+([[?Lost-In-Crowd]], [[?Obscurity-By-Design]]) by rendering the provision of information according to their own
+preferences as easy as possible ([[?Obfuscation]]).
+
+<div class="practice">
+  <span class="practicelab" id="principle-support-choosing-info">
+    APIs should be designed such that data returned through an API does not assert a fact or make a
+    promise on the user's behalf about the user or their environment.
+  </span>
+</div>
+
+Instead, the API could indicate a [=person=]'s preference, a [=person=]'s chosen identity, a
+[=person=]'s query or interest, or a [=person=]'s selected communication style.
+
+In any rare instances when an API must be defined as returning true current values, users may still
+configure their agents to respond with other information, for reasons including testing, auditing or
+mitigating forms of data collection, including <a>browser fingerprinting</a>. In some cases, sites
+do not abide by data minimisation principles and request more information than they require. This
+principle supports [=people=] in defending under such cases.
+
+Sites should include deception in their threat modeling and not assume that Web platform APIs
+provide any guarantees of consistency, currency or correctness about the user. People often have
+control of the devices and software they use to interact with web sites. In response to site
+requests, people may provide information that may be deceptive for a variety of reasons, including
+both malice and self-protection.
+
+Examples of ways in which a [=user agent=] may support this principle include:
+* Generating domain-specific email addresses or other directed identifiers so that [=people=] can log into the
+  site without becoming recognisable across contexts.
+* Offering the option to generate geolocation and accelerometry data with parameters specified by the [=user=].
+* Uploading a stored video stream in response to a camera prompt.
+* Automatically granting or denying permission prompts based on user configuration.
+* Filtering outbound network traffic so as to replace specific strings (hashed email, navigation identifiers, consent strings).
 
 <!--
 # EVERYTHING BELOW THIS LINE IS MATERIAL
--- a/index.html
+++ b/index.html
@@ -1605,7 +1605,9 @@ cases, the absence of data — when a [=person=] refuses to provide some informa
 or revealing. Additionally, APIs can be defined or implemented in rigid ways that can prevent people from
 accessing useful functionality. For example, I might want to look for restaurants in a city I will be visiting
 this weekend, but if my geolocation is forcefully set to match my GPS, a restaurant-finding
-site might only allow searches in my current location.
+site might only allow searches in my current location. In other cases, sites do not abide by data
+minimisation principles and request more information than they require. This principle supports
+[=people=] in minimising their own data.
 
 [=User agents=] should make it simple for [=people=] to [present the identity they wish
 to](#principle-identity-per-context) and to provide information about themselves or their devices in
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1498,6 +1515,60 @@
 
 <aside class="issue">Some services have the user pay for their use in data. These services aren't necessarily retaliating by denying their services to users who refuse to pay with data, but the details are more complex than we've had time to write.</aside>
 
+## Support Choosing Which Information to Present {#support-choosing-info}
+
+<div class="practice">
+  <span class="practicelab" id="principle-support-choosing-info">
+    [=User agents=] should support [=people=] in choosing which information they provide to [=parties=] that
+    request it, up to and including allowing for the provision of information that may not be factually
+    correct.
```
jyasskin, line 1524:
```suggestion
    request it, up to and including allowing users to provide incorrect information.
```

index.html
```diff
@@ -1498,6 +1515,60 @@
...
+</div>
+
+[=Parties=] can invest time and energy into automating ways of gathering [=data=] from [=people=] and can
+design their products in ways that make it a lot easier for [=people=] to disclose information than not, whereas
+[=people=] typically have to manually wade through options, repeated prompts, and deceptive patterns. In many
+cases, the absence of data — when a [=person=] refuses to provide some information — can also be identifying
+or revealing. Additionally, APIs can be defined or implemented in rigid ways that can prevent people from
+accessing useful functionality (eg. if my geolocation is forcefully set to match my GPS, a restaurant-finding
+site might restrict me to my current location when I want to look for restaurants in a city I will be visiting
+this weekend).
```
jyasskin, line 1535: I stumbled over the end of this sentence. Maybe:
```suggestion
or revealing. Additionally, APIs can be defined or implemented in rigid ways that can prevent people from
accessing useful functionality. For example, I might want to look for restaurants in a city I will be visiting
this weekend, but if my geolocation is forcefully set to match my GPS, a restaurant-finding
site might only allow searches in my current location.
```

index.html
```diff
@@ -1498,6 +1515,60 @@
...
+
+[=User agents=] should make it simple for [=people=] to present the identity they wish to and to provide information
```
jyasskin, line 1537: "present the identify they wish to" is already covered by https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/pull/191.html#principle-identity-per-context; do we need to repeat it here? Or we could link to that principle?

```suggestion
[=User agents=] should make it simple for [=people=] to [present the identity they wish to](#principle-identity-per-context) and to provide information
```

index.html
```diff
@@ -1498,6 +1515,60 @@
...
+    APIs should be designed such that data returned through an API does not assert a fact or make a
+    promise on the user's behalf about the user or their environment.
+  </span>
+</div>
+
+Instead, the API could indicate a [=person=]'s preference, a [=person=]'s chosen identity, a
+[=person=]'s query or interest, or a [=person=]'s selected communication style. 
+
+In any rare instances when an API must be defined as returning true current values, users may still
+configure their agents to respond with other information, for reasons including testing, auditing or
```
jyasskin, line 1553:
```suggestion
In any rare instances when an API must be defined as returning true current values, users may still need to
configure their agents to respond with other information, for reasons including testing, auditing or
```
?

index.html
```diff
@@ -1539,28 +1539,35 @@
...
+Instead, the API could indicate a [=person=]'s preference, a [=person=]'s chosen identity, a
+[=person=]'s query or interest, or a [=person=]'s selected communication style. 
 
-In some cases, sites do not abide by data minimisation principles and request more information than they realistically
-require. This principle supports [=people=] in defending under such cases.
+In any rare instances when an API must be defined as returning true current values, users may still
+configure their agents to respond with other information, for reasons including testing, auditing or
+mitigating forms of data collection, including <a>browser fingerprinting</a>. In some cases, sites
+do not abide by data minimisation principles and request more information than they require. This
+principle supports [=people=] in defending under such cases.
```
samuelweiler, line 1556: s/defending under such cases/defending 'against' that behavior/ ?

index.html
npdoty: +1 for linking to that existing principle

index.html
npdoty: "arbitrary" (suggested by Pete, and liked by several)

jyasskin:
```suggestion
    request it, up to and including allowing users to provide arbitrary information.
```

index.html
```diff
@@ -1498,6 +1515,60 @@
...
+
+Sites should include deception in their threat modeling and not assume that Web platform APIs
+provide any guarantees of consistency, currency or correctness about the user. People often have
+control of the devices and software they use to interact with web sites. In response to site
+requests, people may provide information that may be deceptive for a variety of reasons, including
+both malice and self-protection.
```
jyasskin, line 1562: We'll reverse these two paragraphs so that the "users may still configure their agents" bit is clearly advice for site owners rather than a statement about how all APIs actually work.

index.html
jyasskin: I misinterpreted the intent for this sentence.

index.html
```diff
@@ -1498,6 +1515,60 @@
...
+In any rare instances when an API must be defined as returning true current values, users may still
+configure their agents to respond with other information, for reasons including testing, auditing or
+mitigating forms of data collection, including <a>browser fingerprinting</a>. In some cases, sites
+do not abide by data minimisation principles and request more information than they require. This
+principle supports [=people=] in defending under such cases.
+
+Sites should include deception in their threat modeling and not assume that Web platform APIs
+provide any guarantees of consistency, currency or correctness about the user. People often have
+control of the devices and software they use to interact with web sites. In response to site
+requests, people may provide information that may be deceptive for a variety of reasons, including
```
pes10k, line 1561: I suggest changing

> people may provide information that may be deceptive for a variety of reasons, including…

to 

> people may arbitrarily modify or select the information they provide for a variety of reasons, including…

index.html
```diff
@@ -1498,6 +1515,60 @@
...
+Instead, the API could indicate a [=person=]'s preference, a [=person=]'s chosen identity, a
+[=person=]'s query or interest, or a [=person=]'s selected communication style. 
+
+In any rare instances when an API must be defined as returning true current values, users may still
+configure their agents to respond with other information, for reasons including testing, auditing or
+mitigating forms of data collection, including <a>browser fingerprinting</a>. In some cases, sites
+do not abide by data minimisation principles and request more information than they require. This
+principle supports [=people=] in defending under such cases.
+
+Sites should include deception in their threat modeling and not assume that Web platform APIs
```
pes10k, line 1558: I suggest changing 

> Sites should include deception in their threat modeling and not assume that Web platform APIs…

to 

> Sites should expect and accommodate users arbitrarily changing or modifying the responses returned by Web platform APIs, and not assume that the Web platform APIs…

npdoty: Oh, this is the one place where I would prefer to keep "deception" because it is about threat modeling from the perspective of the site operator.

index.html
```diff
@@ -1572,6 +1589,66 @@
...
+For example, a [=user agent=] might support this principle by:
+
+* Generating domain-specific email addresses or other directed identifiers so that [=people=] can
+  log into the site without becoming recognisable across contexts.
+* Offering the option to generate geolocation and accelerometry data with parameters specified by
+  the [=user=].
+* Uploading a stored video stream in response to a camera prompt.
+* Automatically granting or denying permission prompts based on user configuration.
+* Filtering outbound network traffic so as to replace specific strings (hashed email, navigation
+  identifiers, consent strings).
```
jyasskin, line 1636: This doesn't seem connected to the "APIs should be designed" principle, but it is connected to the "support people in choosing" principle. What if we move the two principles to be together at the top of the section, and then the rest of the text can be about both together?

index.html
```diff
@@ -1572,6 +1589,66 @@
...
+or revealing. Additionally, APIs can be defined or implemented in rigid ways that can prevent people from
+accessing useful functionality. For example, I might want to look for restaurants in a city I will be visiting
+this weekend, but if my geolocation is forcefully set to match my GPS, a restaurant-finding
+site might only allow searches in my current location.
+
+[=User agents=] should make it simple for [=people=] to [present the identity they wish
+to](#principle-identity-per-context) and to provide information about themselves or their devices in
+ways that they control. This empowers [=people=] to live in obscurity ([[?Lost-In-Crowd]],
+[[?Obscurity-By-Design]]) by rendering the provision of information according to their own
+preferences as easy as possible ([[?Obfuscation]]).
```
jyasskin, line 1614: I tried to simplify the language, and realized that "provide information about themselves ... in ways that they control" is the same concept as "the provision of information according to their own preferences", so I dropped the second one, and replaced it with a phrase that matches the citation:
```suggestion
ways that they control. This helps [=people=] to live in obscurity ([[?Lost-In-Crowd]],
[[?Obscurity-By-Design]]), including by obfuscating information about themselves ([[?Obfuscation]]).
```

index.html
```diff
@@ -1572,6 +1589,66 @@
...
+* Offering the option to generate geolocation and accelerometry data with parameters specified by
+  the [=user=].
+* Uploading a stored video stream in response to a camera prompt.
+* Automatically granting or denying permission prompts based on user configuration.
+* Filtering outbound network traffic so as to replace specific strings (hashed email, navigation
+  identifiers, consent strings).
+
+</div>
+
+Sites should include deception in their threat modeling and not assume that Web platform APIs
```
jyasskin, line 1640: Is this another principle, aimed at sites as opposed to UAs or API designers?

index.html
```diff
@@ -1572,6 +1589,66 @@
...
+provide any guarantees of consistency, currency, or correctness about the user. People often have
+control of the devices and software they use to interact with web sites. In response to site
+requests, people may people may arbitrarily modify or select the information they provide for a
+variety of reasons, including both malice and self-protection.
+
+In any rare instances when an API must be defined as returning true current values, users may still
+configure their agents to respond with other information, for reasons including testing, auditing or
+mitigating forms of data collection, including <a>browser fingerprinting</a>. In some cases, sites
+do not abide by data minimisation principles and request more information than they require. This
+principle supports [=people=] in defending against that behavior.
```
jyasskin, line 1650: This last sentence goes back to the user-focused principles above (instead of the site-focused principle we started the paragraph with). Maybe we should move it up?

jyasskin (commented):
I had a couple more rearranging ideas that I wanted to run by the group rather than doing them unilaterally. I could also send another PR with these.

index.html
```diff
@@ -1572,6 +1589,66 @@
...
+
+[=User agents=] should make it simple for [=people=] to [present the identity they wish
+to](#principle-identity-per-context) and to provide information about themselves or their devices in
+ways that they control. This empowers [=people=] to live in obscurity ([[?Lost-In-Crowd]],
+[[?Obscurity-By-Design]]) by rendering the provision of information according to their own
+preferences as easy as possible ([[?Obfuscation]]).
+
+<div class="practice">
+  <span class="practicelab" id="principle-no-facts-or-promises">
+    APIs should be designed such that data returned through an API does not assert a fact or make a
```
kdenhartog, line 1618:
```suggestion
    By default, APIs should be designed such that data returned through an API does not assert a fact or make a
```

Thinking this may highlight a nuance I was trying to call out with #191 originally. Which is to say the UA should assume that the user wants the more privacy preserving solution until the user decides to configure the UA to reveal more information about it.

index.html
npdoty: It's certainly site-focused advice. I'm not sure we have marked that as formal principles before. I would be fine either way.

index.html
npdoty: I also found this example a little confusing, so I think it would be fine to leave out.

I don't think the principles need to be headings that all following text is directly part of. That sites should plan for user choice and including deception then leads to a variety of examples.

index.html
npdoty: This text is aimed at the audience of designers of new APIs, so I think "by default" might get confused here (users may have default settings, but there's not as clear a concept of "default" in designing of new standards, except to give general advice, which is what "should be designed" means!).

index.html
pes10k: okie im fine with that

index.html
kdenhartog: Ok - that makes sense and I agree this modification doesn't quite align with the intent then. Hmm I'll have to find a better place for this to go then.

Issue #188, opened 2022-10-26
address privacy threat of harassment


npdoty:
Currently an empty placeholder section in the draft.

torgo:
we agreed we can close this based on #195.

PR #195, opened 2022-11-16
initial text and principle around harassment

npdoty:
initial pass at #188

### DIFF
--- a/index.html
+++ b/index.html
@@ -239,6 +239,17 @@
           href: 'https://www.oecd.org/sti/ieconomy/oecdguidelinesontheprotectionofprivacyandtransborderflowsofpersonaldata.htm',
           publisher: 'OECD',
         },
+        'PEN-Harassment': {
+          href: 'https://onlineharassmentfieldmanual.pen.org/defining-online-harassment-a-glossary-of-terms/',
+          title: 'Online Harassment Field Manual',
+          publisher: 'PEN America',
+        },
+        'PEW-Harassment': {
+          title: 'The State of Online Harassment',
+          publisher: 'Pew Research Center',
+          date: 'January 2021',
+          href: 'https://www.pewresearch.org/internet/2021/01/13/the-state-of-online-harassment/'
+        },
         'Phone-On-Feminism': {
           title: 'This is your phone on feminism',
           href: 'https://conversationalist.org/2019/09/13/feminism-explains-our-toxic-relationships-with-our-smartphones/',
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1430,9 +1441,48 @@
...
 
-<aside class="issue">
-  This section has not been written.
-</aside>
+Online <dfn>harassment</dfn> is the "pervasive or severe targeting of an individual or group online
+through harmful behavior" [[PEN-Harassment]]. Harassment is a prevalent problem on the Web,
+particularly via social media. Harassment may be more severe and its consequences more impactful for
+LGBTQ people and women [[?PEW-Harassment]].
+
+<a>Harassment</a> is both a violation of privacy itself and can be a consequence of other violations
```
darobin, line 1449:
```suggestion
[=Harassment=] is both a violation of privacy itself and can be a consequence of other violations
```

darobin: This is just a MD nitpick.

darobin, line 1449:
```suggestion
<a>Harassment</a> is both a violation of privacy itself and can be magnified or facilitated by other violations
```

I don't think that it's really a consequence of other privacy violations but they can contribute.

index.html
```diff
@@ -1430,9 +1441,48 @@
...
+of privacy. 
+
+Abusive online behavior may include: sending <a href="#unwanted-information">unwanted
+information</a>; directing others to contact or bother a person ("dogpiling"); disclosing sensitive
```
darobin, line 1453: Nitpick but "unwanted information" should be a `dfn` and this could reference it.

index.html
```diff
@@ -1430,9 +1441,48 @@
...
+information about a person; posting false information about a person; impersonating a person;
+insults; threats; and hateful or demeaning speech.
+
+Disclosure of contact information can be used, including by a separate attacker, to send often
+persistent unwanted information that amounts to harassment. Disclosure of location information can
+be used, including by a separate attacker, to intrude on a person's physical safety or space.
```
darobin, line 1459: Might it be worth mentioning doxing and swatting?

index.html
```diff
@@ -1430,9 +1441,48 @@
...
+
+Mitigations for harassment include but extend beyond mitigations for unwanted information and other
+privacy principles. Harassment can include harmful activity not solely directly communicated to the
+target of harassment.
+
+<div class="practice">
+  <p>
+    <span class="practicelab" id="abuse-reporting">
+      Systems that allow for communicating on the Web must provide an effective capability to report
+      abuse.
```
darobin, line 1469: Reporting _and_ enforcement, no?

darobin (changes requested):
Overall I think it's a valuable addition, but there are several small details and formatting issues to address.

index.html
```diff
@@ -1430,9 +1441,48 @@
...
+
+
+<div class="note">
+  Effective reporting is likely to require:
+
+  * standardized mechanisms to identify abuse reporting contacts * visible, usable ways provided by
+  sites and user agents to report abuse * identifiers to refer to senders and content * the ability
+  to provide context and explanation of harms * people responsible for promptly responding to
+  reports * tools for pooling mitigation information (see Unwanted information, below)  
+</div>
```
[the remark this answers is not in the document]
npdoty: thanks, that's what I get for not checking it again after that last moment text-wrapping

index.html
npdoty: That's a good topic for broader discussion. I do think some enforcement is needed in order for the reporting abuse mitigation to be effective. I'm less clear on how specifically we can give guidance on that (or how often that would be in the standards or web implementations discussed at w3c), but I would welcome a text suggestion on that.

darobin: I don't think that we could give good guidance on enforcement at this stage, but I think that keeping enforcement implicit behind "effective" is a bit too indirect. Maybe "must provide an effective capability to report abuse coupled with forceful enforcement."

index.html
npdoty: added an explicit doxxing reference
swatting is included in the PEN reference, and would certainly be an example of physical intrusion, but I don't know if we need an explicit reference to that as well

index.html
```diff
@@ -1430,9 +1441,50 @@
...
+
+Abusive online behavior may include: sending [=unwanted information=]; directing others to contact
+or bother a person ("dogpiling"); disclosing sensitive information about a person; posting false
+information about a person; impersonating a person; insults; threats; and hateful or demeaning
+speech.
+
+Disclosure of identifying or contact information (including "doxxing") can be used, including by a
+separate attacker, to send often persistent unwanted information that amounts to harassment.
+Disclosure of location information can be used, including by a separate attacker, to intrude on a
+person's physical safety or space.
```
jyasskin, line 1460: What's a "separate attacker"?

index.html
```diff
@@ -1430,9 +1441,50 @@
 
 ## Harassment
 
-<aside class="issue">
-  This section has not been written.
-</aside>
+Online <dfn>harassment</dfn> is the "pervasive or severe targeting of an individual or group online
+through harmful behavior" [[PEN-Harassment]]. Harassment is a prevalent problem on the Web,
+particularly via social media. Harassment may be more severe and its consequences more impactful for
+LGBTQ people and women [[?PEW-Harassment]].
```
jyasskin, line 1447: We have to mention non-white races if we list groups, and we'll probably get in trouble for missing other disadvantaged groups. How about:

```suggestion
particularly via social media. Harassment may be more severe and its consequences more impactful for
people who are disadvantaged in society.
```

index.html
```diff
@@ -1430,9 +1441,50 @@
...
+information about a person; impersonating a person; insults; threats; and hateful or demeaning
+speech.
+
+Disclosure of identifying or contact information (including "doxxing") can be used, including by a
+separate attacker, to send often persistent unwanted information that amounts to harassment.
+Disclosure of location information can be used, including by a separate attacker, to intrude on a
+person's physical safety or space.
+
+Mitigations for harassment include but extend beyond mitigations for unwanted information and other
+privacy principles. Harassment can include harmful activity not solely directly communicated to the
```
jyasskin, line 1463: "solely directly communicated" is a bit obscure. "... with a wider distribution than just the target ..."?

index.html
torgo: We agreed to leave enforcement out for now in the call today. Can revisit in future.

jyasskin (approved):

torgo (approved):
lgtm

PR #215, opened 2023-01-25
Add text on aggregated data

dmarti:
From meeting discussion: https://github.com/w3ctag/privacy-principles/blob/main/meetings/2023-01-18-minutes.md#contributing-to-an-aggregate

PR #216, opened 2023-01-25
Revisions and concisions around ancillary data use

pes10k:
Comments and suggestions in response to https://github.com/w3ctag/privacy-principles/issues/165

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/pull/216.html#ancillary-uses" title="Last updated on Feb 23, 2023, 12:45 AM UTC (1b585f1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/216/a1d8862...1b585f1.html" title="Last updated on Feb 23, 2023, 12:45 AM UTC (1b585f1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1238,27 +1238,26 @@ Note that this principle was further explored in an earlier TAG draft on [[[Data
 <div class="practice">
 <span class="practicelab">In maintaining duties of [=duty of protection|protection=], [=duty of discretion|discretion=] and [=duty of loyalty|loyalty=], user
 agents should take care that, when sharing data that isn't needed to satisfy a user's immediate goals, such information sharing
-aligns with the user's wishes and interests.</span>
+aligns with the user's intents and interests.</span>
 </div>
 
-On the Web, user agents and sites often communicate data outside what is needed to load and display a
-page. These include:
+On the Web, user agents and sites can communicate data outside what is needed
+to load and display a page, data not needed to accomplish the user's intent on a site.
+This category of data is called <dfn data-lt="ancillary data">ancillary data</dfn>,
+and these uses are <dfn data-lt="ancillary use">ancillary uses</dfn> of data.
 
-<dfn>Telemetry and analytics</dfn>, including browser telemetry and site telemetry, are common
-categories of <dfn data-lt="ancillary use">ancillary uses</dfn> of data.
-
-In some cases, aggregate data collection mechanisms
-can be used in place of other
-more intrusive or costly data collection mechanisms.
-
-Different users will want to share different kinds and amounts of data with websites.
-In some case, users might reasonably anticipate of some data being collected
-for an [=ancillary uses=] because of a contextual norm the user approves of. In such cases,
+Different users will want to share different kinds and amounts of ancillary data
+with websites, including possibly no ancillary data.
+In some case, users might reasonably anticipate some data being collected
+for an [=ancillary uses=], possibly because of a contextual norm the user approves of. In such cases,
 users are identifying such ancillary uses as aligned with their own goals.
 
-Aggregated or [=de-identified=] data may contribute to a collective benefit in a way
+Aggregating or [=de-identification|de-identified=] of data may make users
+interested in sharing ancillary data in cases were the user was
+otherwise not interested. These techniques may be especially useful and important
+when ancillary data contributes to a collective benefit in a way
 that reduces privacy threats to individuals (see <a href="#principle-collective-privacy">collective
-privacy</a>). This may make users more interested in contributing their data to aggregates.
+privacy</a>).
 
 <aside class="example">
   Privacy-preserving measurement techniques may be used for aggregate calculations while minimizing
--- a/index.html
+++ b/index.html
@@ -1294,6 +1294,14 @@ communicate common user preferences (such as a [=global opt-out=]).
   choices.</span>
 </div>
 
+<aside class="example">
+  Sites and browsers wish to collect telemetry data to determine how frequently features are used or
+  to debug breakages, but the user agent does not want to burden the user with frequent consent
+  requests. A browser could use a first-run dialog to ask the user whether they generally support
+  sharing data to find bugs and improve the Web software they use, and then enable or disable
+  telemetry and reporting APIs based on the user's choice.
+</aside>
+
 Data exposed for [=ancillary uses=] including telemetry and analytics may
 often reveal characteristics of user configuration, device, environment, or behavior that could be
 used as part of <a>browser fingerprinting</a> to identify users across sites. Revealing user
--- a/index.html
+++ b/index.html
@@ -1264,7 +1264,7 @@ privacy</a>).
   Privacy-preserving measurement techniques may be used for aggregate calculations while minimizing
   the number of actors that have access to personal data about many individual people. Encryption and
   privacy-preserving proxies may minimize the number of actors that have access to personal data or
-  to hide the contents of personal data. But even
+  hide the contents of personal data. But even
   with those protections, some people may prefer not to participate in some kinds of measurement.
 
   Ongoing work on privacy-preserving technologies in the <abbr title="Internet Engineering Task
--- a/index.html
+++ b/index.html
@@ -1248,17 +1248,17 @@ goals. These uses are known as <dfn data-lt="ancillary use">ancillary uses</dfn>
 and data that is primarily useful for [=ancillary uses=] is <dfn>ancillary data</dfn>.
 
 <aside class="example">
-  Some examples of ancillary data include data used for browser telemetry, site telemetry,
+  Some examples of [=ancillary data=] include data used for browser telemetry, site telemetry,
   performance measurements, and software updates.
 </aside>
 
-Different users will want to share different kinds and amounts of ancillary data
-with websites, including possibly no ancillary data.
+Different users will want to share different kinds and amounts of [=ancillary data=]
+with websites, including possibly no [=ancillary data=].
 
 Aggregation or [=de-identified|de-identification=] of data may make users
-interested in sharing ancillary data in cases were the user was
+interested in sharing [=ancillary data=] in cases were the user was
 otherwise not interested. These techniques may be especially useful and important
-when ancillary data contributes to a collective benefit in a way
+when [=ancillary data=] contributes to a collective benefit in a way
 that reduces privacy threats to individuals (see <a href="#principle-collective-privacy">collective
 privacy</a>).
 
--- a/index.html
+++ b/index.html
@@ -1236,9 +1236,11 @@ Note that this principle was further explored in an earlier TAG draft on [[[Data
 </aside>
 
 <div class="practice">
-<span class="practicelab">In maintaining duties of [=duty of protection|protection=], [=duty of discretion|discretion=] and [=duty of loyalty|loyalty=], user
-agents should take care that, when sharing data that isn't needed to satisfy a user's immediate goals, such information sharing
-aligns with the user's intents and interests.</span>
+<span class="practicelab">In maintaining duties of [=duty of
+protection|protection=], [=duty of discretion|discretion=] and [=duty of
+loyalty|loyalty=], user agents should share data only when it either is needed
+to satisfy a user's immediate goals or aligns with the user's wishes and
+interests.</span>
 </div>
 
 Websites sometimes use data in ways that aren't needed for the user's immediate
--- a/index.html
+++ b/index.html
@@ -1241,10 +1241,9 @@ agents should take care that, when sharing data that isn't needed to satisfy a u
 aligns with the user's intents and interests.</span>
 </div>
 
-On the Web, user agents and sites can communicate data outside what is needed
-to load and display a page, data not needed to accomplish the user's intent on a site.
-This category of data is called <dfn data-lt="ancillary data">ancillary data</dfn>,
-and these uses are <dfn data-lt="ancillary use">ancillary uses</dfn> of data.
+Websites sometimes use data in ways that aren't needed for the user's immediate
+goals. These uses are known as <dfn data-lt="ancillary use">ancillary uses</dfn>,
+and data that is primarily useful for [=ancillary uses=] is <dfn>ancillary data</dfn>.
 
 <aside class="example">
   Some examples of ancillary data include browser telemetry, site telemetry,
@@ -1253,11 +1252,8 @@ and these uses are <dfn data-lt="ancillary use">ancillary uses</dfn> of data.
 
 Different users will want to share different kinds and amounts of ancillary data
 with websites, including possibly no ancillary data.
-In some case, users might reasonably anticipate some data being collected
-for an [=ancillary uses=], possibly because of a contextual norm the user approves of. In such cases,
-users are identifying such ancillary uses as aligned with their own goals.
 
-Aggregating or [=de-identification|de-identified=] of data may make users
+Aggregation or [=de-identified|de-identification=] of data may make users
 interested in sharing ancillary data in cases were the user was
 otherwise not interested. These techniques may be especially useful and important
 when ancillary data contributes to a collective benefit in a way
@@ -1285,8 +1281,8 @@ privacy</a>).
   goals and preferences about use of data about them.</span>
 </div>
 
-Agents should be particularly careful, cautious, and aggresive in applying data
-minimization to data intended for [=ancillary uses=]. To facilitate site
+Agents should aggressively <a href="#data-minimization">minimize</a> [=ancillary data=].
+To facilitate site
 understanding of user preferences, user agents can provide browser-configurable signals to directly
 communicate common user preferences (such as a [=global opt-out=]).
 
--- a/index.html
+++ b/index.html
@@ -1288,11 +1288,11 @@ communicate common user preferences (such as a [=global opt-out=]).
 
 <div class="practice">
   <span class="practicelab">Specifications that define functionality for telemetry and analytics
-  should explicitly note the [=telemetry and analytics|analytics=] use to facilitate modal or general user
+  should explicitly note the telemetry and analytics use to facilitate modal or general user
   choices.</span>
 </div>
 
-Data exposed for [=ancillary uses=] including [=telemetry and analytics|telemetry=] may
+Data exposed for [=ancillary uses=] including telemetry and analytics may
 often reveal characteristics of user configuration, device, environment, or behavior that could be
 used as part of <a>browser fingerprinting</a> to identify users across sites. Revealing user
 preferences or other heuristics in providing or disabling functionality could also contribute to a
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1235,53 +1235,36 @@
   uses</a> for ongoing discussion and iteration on this principle.
 </aside>
 
-In maintaining duties of [=duty of protection|protection=], [=duty of discretion|discretion=] and [=duty of loyalty|loyalty=], user
-agents should be conservative when sharing data that isn't needed to satisfy a user's immediate goals.
+<div class="practice">
+<span class="practicelab">In maintaining duties of [=duty of protection|protection=], [=duty of discretion|discretion=] and [=duty of loyalty|loyalty=], user
+agents should take care that, when sharing data that isn't needed to satisfy a user's immediate goals, such information sharing
```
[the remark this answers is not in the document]
pes10k: I dig it! I'll update the PR accordingly after the call today (just to batch the updates)

index.html
```diff
@@ -1298,10 +1281,8 @@
   goals and preferences about use of data about them.</span>
 </div>
 
-Agents should be careful, cautious, and conservative in applying data minimization to any
-[=ancillary use=], but also not burden the user with additional [=privacy
-labor=]. To that end, user agents may employ user research, solicitation of general preferences, and
-heuristics about sensitivity of data or trust in a particular context. To facilitate site
+Agents should be particularly careful, cautious, and aggresive in applying data
```
samuelweiler, line 1284: this change LGTM

index.html
```diff
@@ -1235,53 +1235,36 @@
...
+Aggregated or [=de-identified=] data may contribute to a collective benefit in a way
+that reduces privacy threats to individuals (see <a href="#principle-collective-privacy">collective
+privacy</a>). This may make users more interested in contributing their data to aggregates.
 
 <aside class="example">
   Privacy-preserving measurement techniques may be used for aggregate calculations while minimizing
   the number of actors that have access to personal data about many individual people. Encryption and
-  privacy-preserving proxies may be used to minimize the number of actors that have access to personal data or
-  to hide the contents of personal data used for telemetry or other kinds of measurement. But even
+  privacy-preserving proxies may minimize the number of actors that have access to personal data or
```
samuelweiler, line 1266: LGTM

index.html
```diff
@@ -1235,53 +1235,36 @@
...
-some people would reasonably anticipate sharing such data or identify that use as a
-commonly-accepted contextual norm and some people would be willing to participate (for example,
-might have explicitly decided to contribute if they had the time and understanding to consider
-details about such a use) and some people would identify those uses as supporting their own goals.
-
-Aggregated or [=de-identified=] data is often more likely to be
-acceptable in these senses and often still contributes to the collective benefit while minimizing
-privacy threats to a particular individual (see <a href="#principle-collective-privacy">collective
-privacy</a>). But some people might object anyway, in particular instances or in general.
+page. These include:
```
jyasskin, line 1245: "These include:" is out of place. There's no list after this.

index.html
```diff
@@ -1235,53 +1235,36 @@
...
+more intrusive or costly data collection mechanisms.
+
+Different users will want to share different kinds and amounts of data with websites.
+In some case, users might reasonably anticipate of some data being collected
+for an [=ancillary uses=] because of a contextual norm the user approves of. In such cases,
+users are identifying such ancillary uses as aligned with their own goals.
+
+Aggregated or [=de-identified=] data may contribute to a collective benefit in a way
+that reduces privacy threats to individuals (see <a href="#principle-collective-privacy">collective
+privacy</a>). This may make users more interested in contributing their data to aggregates.
```
jyasskin, line 1262: Suggested addition from @dmarti's #215.

```suggestion

A [=user agent=] must act in the interest of its user when reporting aggregated data and should
not use the fact that data is aggregated as an excuse to share more data than is in that user's interest.

```

I haven't double-checked that this is the best place to insert this in the new flow.

index.html
npdoty: +1

index.html
npdoty, line 1284: Cautious and aggressive seem contradictory, so I'm not sure how the reader will interpret it. If what we mean is that the agent should have a tendency towards minimizing the data when it's an ancillary use, I think "conservative" covers that. "Conservative" was also removed from the introductory text at the beginning, though I'm not quite sure why. "Conservative" was in Pete's text originally, which was why we had kept it in this section.

"Particularly" implies that this agents should do this in particular, but I don't see what the more general case is that we're contrasting against. The other principles in this section and the other principles in this document seem equally relevant and important.

index.html
```diff
@@ -1311,19 +1292,6 @@
   choices.</span>
 </div>
 
-<aside class="example">
-  Sites and browsers wish to collect telemetry data to determine how frequently features are used or
-  to debug breakages, but the user agent does not want to burden the user with frequent consent
-  requests. A browser could use a first-run dialog to ask the user whether they generally support
-  sharing data to find bugs and improve the Web software they use, and then enable or disable
-  telemetry and reporting APIs based on the user's choice.
-</aside>
```
npdoty, line 1320: When the group was discussing this topic, this example had what I believe was strongest consensus for it, in documenting how we wanted to see general preferences solicited in a way that would avoid unnecessary prompts and labor.

index.html
```diff
@@ -1235,53 +1235,36 @@
...
-details about such a use) and some people would identify those uses as supporting their own goals.
-
-Aggregated or [=de-identified=] data is often more likely to be
-acceptable in these senses and often still contributes to the collective benefit while minimizing
-privacy threats to a particular individual (see <a href="#principle-collective-privacy">collective
-privacy</a>). But some people might object anyway, in particular instances or in general.
+page. These include:
+
+<dfn>Telemetry and analytics</dfn>, including browser telemetry and site telemetry, are common
+categories of <dfn data-lt="ancillary use">ancillary uses</dfn> of data.
```
npdoty, line 1248: This makes the definition of ancillary uses less clear because there's only one category of example now, and that example is in between the general definition and the defined term. Does "ancillary uses" still include non-telemetry uses?

index.html
```diff
@@ -1235,53 +1235,36 @@
...
+
+In some cases, aggregate data collection mechanisms
+can be used in place of other
+more intrusive or costly data collection mechanisms.
+
+Different users will want to share different kinds and amounts of data with websites.
+In some case, users might reasonably anticipate of some data being collected
+for an [=ancillary uses=] because of a contextual norm the user approves of. In such cases,
+users are identifying such ancillary uses as aligned with their own goals.
```
npdoty, line 1257: This seems to imply that there is a single way to determine whether an ancillary use is aligned with the user's goals, which is: the user reasonably anticipates it; anticipates it because of a contextual norm; and approves of the contextual norm. I don't think that matches with the group's understanding, and I don't think that would include most telemetry cases, for example.

index.html
```diff
@@ -1235,53 +1235,36 @@
...
-privacy threats to a particular individual (see <a href="#principle-collective-privacy">collective
-privacy</a>). But some people might object anyway, in particular instances or in general.
+page. These include:
+
+<dfn>Telemetry and analytics</dfn>, including browser telemetry and site telemetry, are common
+categories of <dfn data-lt="ancillary use">ancillary uses</dfn> of data.
+
+In some cases, aggregate data collection mechanisms
+can be used in place of other
+more intrusive or costly data collection mechanisms.
```
npdoty, line 1252: This is true, but reads as if it's a recommendation rather than descriptive text, and now only considers aggregate mechanisms. We could just remove this paragraph if we don't want to have descriptive text. Including descriptive text but only referring to aggregate mechanisms when there are also non-aggregate data collection mechanisms used in this way would likely confuse the reader.

index.html
```diff
@@ -1235,53 +1235,36 @@
...
-cluster of ancillary uses.
-
-In some cases, mechanisms for sharing data are provided in order to minimize some other data
-collection mechanism identified as more intrusive or costly.
-
-Different users will want to share different kinds and amounts of data with websites. For some uses,
-some people would reasonably anticipate sharing such data or identify that use as a
-commonly-accepted contextual norm and some people would be willing to participate (for example,
-might have explicitly decided to contribute if they had the time and understanding to consider
-details about such a use) and some people would identify those uses as supporting their own goals.
```
npdoty, line 1273: Removing these different ways the group identified as how user interests might vary seems to remove some important conclusions from the group. But if we are going to remove this descriptive text, I think it would be better to just remove it rather than replacing it with a single example and implying that only one way counts.

index.html
npdoty: Removing the privacy labor point removes the conclusion the group had that we didn't want our recommendations to lead to a shift in responsibility of the labor to the user with more prompts.

index.html
```diff
@@ -1235,53 +1235,36 @@
   uses</a> for ongoing discussion and iteration on this principle.
 </aside>
 
-In maintaining duties of [=duty of protection|protection=], [=duty of discretion|discretion=] and [=duty of loyalty|loyalty=], user
-agents should be conservative when sharing data that isn't needed to satisfy a user's immediate goals.
+<div class="practice">
+<span class="practicelab">In maintaining duties of [=duty of protection|protection=], [=duty of discretion|discretion=] and [=duty of loyalty|loyalty=], user
+agents should take care that, when sharing data that isn't needed to satisfy a user's immediate goals, such information sharing
+aligns with the user's wishes and interests.</span>
```
npdoty, line 1241: I'm not sure making this into a principle will be very actionable, but no objection.

There are no other references or definitions of user wishes, so maybe we could stick to interests or aligned with goals, unless we have an additional distinct meaning.

index.html
npdoty: Does this add anything to our definition of the duty of loyalty? I'm not sure we should add "don't violate the duty of loyalty by making up excuses" to each section.

index.html
npdoty: "intents and interests" seems promising for the group

index.html
```diff
@@ -1235,53 +1235,35 @@
...
+<div class="practice">
+<span class="practicelab">In maintaining duties of [=duty of protection|protection=], [=duty of discretion|discretion=] and [=duty of loyalty|loyalty=], user
+agents should take care that, when sharing data that isn't needed to satisfy a user's immediate goals, such information sharing
+aligns with the user's intents and interests.</span>
+</div>
+
+On the Web, user agents and sites can communicate data outside what is needed
+to load and display a page, data not needed to accomplish the user's intent on a site.
+This category of data is called <dfn data-lt="ancillary data">ancillary data</dfn>,
+and these uses are <dfn data-lt="ancillary use">ancillary uses</dfn> of data.
```
jyasskin, line 1247:
```suggestion
Websites sometimes use data in ways that aren't needed for the user's immediate
goals. These uses are known as <dfn data-lt="ancillary use">ancillary uses</dfn>,
and data that is only useful for [=ancillary uses=] is <dfn>ancillary data</dfn>
```

jyasskin: Or maybe "data that is primarily useful for ancillary uses":

```suggestion
Websites sometimes use data in ways that aren't needed for the user's immediate
goals. These uses are known as <dfn data-lt="ancillary use">ancillary uses</dfn>,
and data that is primarily useful for [=ancillary uses=] is <dfn>ancillary data</dfn>.
```

index.html
pes10k: interested -> comfortable

index.html
```diff
@@ -1235,53 +1235,35 @@
...
+
+Different users will want to share different kinds and amounts of ancillary data
+with websites, including possibly no ancillary data.
+In some case, users might reasonably anticipate some data being collected
+for an [=ancillary uses=], possibly because of a contextual norm the user approves of. In such cases,
+users are identifying such ancillary uses as aligned with their own goals.
+
+Aggregating or [=de-identification|de-identified=] of data may make users
```
jyasskin, line 1255: Respec syntax and aligning grammar:

```suggestion
Aggregation or [=de-identified|de-identification=] of data may make users
```

index.html
pes10k: delete lines: 1251-1253

index.html
```diff
@@ -1235,53 +1235,35 @@
...
+On the Web, user agents and sites can communicate data outside what is needed
+to load and display a page, data not needed to accomplish the user's intent on a site.
+This category of data is called <dfn data-lt="ancillary data">ancillary data</dfn>,
+and these uses are <dfn data-lt="ancillary use">ancillary uses</dfn> of data.
+
+Different users will want to share different kinds and amounts of ancillary data
+with websites, including possibly no ancillary data.
+In some case, users might reasonably anticipate some data being collected
+for an [=ancillary uses=], possibly because of a contextual norm the user approves of. In such cases,
+users are identifying such ancillary uses as aligned with their own goals.
```
jyasskin, line 1253: We've suggested removing this 

```suggestion
```

We've suggested removing these sentences today. The idea of "if they had the time and understanding to consider details about such a use" is good, but doesn't quite fit here, so we'll keep thinking about places it might fit better, and until then, shorter is better.

index.html
```diff
@@ -1298,10 +1280,8 @@
   goals and preferences about use of data about them.</span>
 </div>
 
-Agents should be careful, cautious, and conservative in applying data minimization to any
-[=ancillary use=], but also not burden the user with additional [=privacy
-labor=]. To that end, user agents may employ user research, solicitation of general preferences, and
-heuristics about sensitivity of data or trust in a particular context. To facilitate site
+Agents should be particularly careful, cautious, and aggresive in applying data
+minimization to data intended for [=ancillary uses=]. To facilitate site
```
jyasskin, line 1284:
```suggestion
Agents should aggressively <a href="#data-minimization">minimize</a> [=ancillary data=].
To facilitate site
```

index.html
```diff
@@ -1235,53 +1235,35 @@
   uses</a> for ongoing discussion and iteration on this principle.
 </aside>
 
-In maintaining duties of [=duty of protection|protection=], [=duty of discretion|discretion=] and [=duty of loyalty|loyalty=], user
-agents should be conservative when sharing data that isn't needed to satisfy a user's immediate goals.
-
-On the Web, user agents and sites often communicate data outside what is needed to load and display a
-page. These <dfn data-lt="ancillary use">ancillary uses</dfn> include:
-
-* browser telemetry, including:
```
pes10k, line 1244: Pete will move these into "example" and use just the highest level in the list

index.html
pes10k: i dig it

index.html
pes10k: i dig it (thanks!)

index.html
pes10k: i dig it (removing it from here, and discussing further if / when there is a suggestion for where it might fit better)

index.html
pes10k: i dig it

index.html
pes10k: I've added the top level examples back into an "example" block, and moved it directly after our informal definitions of ancillary data and ancillary uses of data

torgo (approved):
lgtm let's ship it

index.html
```diff
@@ -1235,53 +1235,36 @@
...
+<div class="practice">
+<span class="practicelab">In maintaining duties of [=duty of protection|protection=], [=duty of discretion|discretion=] and [=duty of loyalty|loyalty=], user
+agents should take care that, when sharing data that isn't needed to satisfy a user's immediate goals, such information sharing
+aligns with the user's intents and interests.</span>
+</div>
+
+Websites sometimes use data in ways that aren't needed for the user's immediate
+goals. These uses are known as <dfn data-lt="ancillary use">ancillary uses</dfn>,
+and data that is primarily useful for [=ancillary uses=] is <dfn>ancillary data</dfn>.
+
```
npdoty, line 1247: I still don't see the benefit from separately defining ancillary data. It will be tricky to determine whether data is *primarily* useful for ancillary uses, and I don't think it would change how we recommend using minimization. Are User-Agent strings primarily useful for ancillary uses, or are they regularly used for both immediate goals and ancillary uses?

index.html
```diff
@@ -1235,53 +1235,36 @@
...
+<aside class="example">
+  Some examples of ancillary data include browser telemetry, site telemetry,
+  performance measurements, and software updates.
+</aside>
```
npdoty, line 1251: These seem like examples of uses, rather than data types.

index.html
```diff
@@ -1298,33 +1281,18 @@
   goals and preferences about use of data about them.</span>
 </div>
 
-Agents should be careful, cautious, and conservative in applying data minimization to any
-[=ancillary use=], but also not burden the user with additional [=privacy
-labor=]. To that end, user agents may employ user research, solicitation of general preferences, and
```
npdoty, line 1303: Removing reference to the concept of privacy labor from this section seems unhelpful. It's been raised regularly, both internal to our group and from external parties. It also explains the reasoning for why we provide examples of general or global user signals.

index.html
jyasskin:
```suggestion

<aside class="issue">
The task force isn't sure we should be defining "ancillary data". If data's rarely clearly ancillary or for immediate goals, then we should reword the text to focus on the uses rather than the data.
</aside>
```

index.html
```diff
@@ -1298,33 +1281,18 @@
...
+To facilitate site
 understanding of user preferences, user agents can provide browser-configurable signals to directly
 communicate common user preferences (such as a [=global opt-out=]).
 
 <div class="practice">
   <span class="practicelab">Specifications that define functionality for telemetry and analytics
-  should explicitly note the [=telemetry and analytics|analytics=] use to facilitate modal or general user
+  should explicitly note the telemetry and analytics use to facilitate modal or general user
   choices.</span>
 </div>
```
jyasskin, line 1294:
```suggestion

<aside class="example">
  Sites and browsers wish to collect telemetry data to determine how frequently features are used or
  to debug breakages, but the user agent does not want to burden the user with frequent consent
  requests. A browser could use a first-run dialog to ask the user whether they generally support
  sharing data to find bugs and improve the Web software they use, and then enable or disable
  telemetry and reporting APIs based on the user's choice.
</aside>

```

index.html
```diff
@@ -1235,53 +1235,36 @@
...
+aligns with the user's intents and interests.</span>
+</div>
+
+Websites sometimes use data in ways that aren't needed for the user's immediate
+goals. These uses are known as <dfn data-lt="ancillary use">ancillary uses</dfn>,
+and data that is primarily useful for [=ancillary uses=] is <dfn>ancillary data</dfn>.
+
+<aside class="example">
+  Some examples of ancillary data include browser telemetry, site telemetry,
+  performance measurements, and software updates.
```
jyasskin, line 1250:
```suggestion
  Some examples of ancillary data include data used for browser telemetry, site telemetry,
  performance measurements, and software updates.
```

PR #227, opened 2023-03-09
Fix #224: Mention harm from partially blocking cross-context recognition.

### LINKED ISSUE #224: Using the privacy argument as an excuse to discriminate
**This is section 1 from the document "Commentary on the draft W3C Privacy Principles" from Movement for An Open Web, attached.**

MOW agrees that privacy should not be used as a weapon by large entities to favour their own businesses through
discriminatory practices by alleging that end users must be protected from the influence of so called third-parties.1

MOW further agrees that larger entities tend to have more power relative to both people and markets, and as such,
must be better protected from abuses they may perpetrate given this situation.

“One of the ways in which the Web serves people is by protecting them in the face of asymmetries of power,
and this includes establishing and enforcing rules to govern the power of data....”

Unfortunately, the document goes on to state the following.

> “Information is power. It can be used to predict and to influence people, as well as to design online spaces
that control people's behaviour. The collection and processing of information in greater volume, with
greater precision and reliability, with increasing interoperability across a growing variety of data types, and
at intensifying speed is leading to an unprecedented concentration of power that threatens private and
public liberties.”2

It is well known that Google and Apple collect and process vast troves of personal data from individuals’ interactions
with rival online businesses via their OS, app stores and browsers.3

However, the document makes an unfounded claim that information used to “predict” behaviour, or provide
desirable information to “influence” behaviour, somehow magically transforms into a force that enables software
to remove free will and all decision-making ability by “control[ing] people’s behavior.”

A position MOW supports is summarized by Cory Doctorow, a leading online digital expert:

> To understand why you shouldn’t worry about mind-control rays — but why you should worry about
surveillance and Big Tech — we must start by unpacking what we mean by “persuasion.”
> Google, Facebook, and other surveillance capitalists promise their customers (the advertisers) that if they
use machine-learning tools trained on unimaginably large data sets of nonconsensually harvested personal
information, they will be able to uncover ways to bypass the rational faculties of the public and direct their
behavior, creating a stream of purchases, votes, and other desired outcomes.
> The impact of dominance far exceeds the impact of manipulation and should be central to our analysis and
any remedies we seek.
> But there’s little evidence that this is happening. Instead, the predictions that surveillance capitalism delivers
to its customers are much less impressive. Rather than finding ways to bypass our rational faculties,
surveillance capitalists like Mark Zuckerberg mostly do one or more of three things:
> Segmenting [while using sensitive data it can be] "seriously creepy. But it’s not mind control....”
> Deception... This is pernicious and difficult — and it’s also the kind of thing the internet can help guard
against by making true information available, especially in a form that exposes the underlying deliberations
among parties with sharply divergent views, such as Wikipedia. But it’s not brainwashing; it’s fraud.”
> Domination. Surveillance capitalism is the result of monopoly. Monopoly is the cause, and surveillance
capitalism and its negative outcomes are the effects of monopoly.... that has allowed companies to grow by
merging with their rivals, buying up their nascent competitors, and expanding to control whole market
verticals. One example of how monopolism aids in persuasion is through dominance: Google makes editorial
decisions about its algorithms that determine the sort order of the responses to our queries.”4

The authors of these Privacy Principles fortunately understand the dangers to individuals of increased centralization
that technical standards can foster, and the benefits provided by competition.

> “While privacy principles are designed to work together and support each other, occasionally a proposal to
improve how a system follows one privacy principle may reduce how well it follows another principle.”5

Thus, when designing standards, we must ensure we understand the risk to

> “a person using an essential service provided by a monopolistic platform — and those where people and
parties are very much on equal footing, or even where the person may have greater power, as is the case
with small businesses operating in a competitive environment.”6

For any policy to have a chance at improving the situation, it is critical to understand both the problem we are
addressing and how well we predict the remedy would address it.

The Principles claim to provide a “trustworthy platform” that supports a “Web for all,” which would require it to
work for individuals and organizations of all sizes with which people interact. Accordingly, if a proposed principle
“only benefits powerful, large entities that control both an implementation and services,” this would be “harmful to
the Web.”7

Accordingly, we must analyse each principle listed as to whether it benefits online organizations of all sizes or (even
unintentionally) favours larger entities and discriminates against smaller ones.

[2022.05.31-MOW-response-to-the-W3C-Privacy-Principles-EB.pdf](https://github.com/w3ctag/privacy-principles/files/10915527/2022.05.31-MOW-response-to-the-W3C-Privacy-Principles-EB.pdf)

jyasskin:
***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/227.html#user-agent-recognition" title="Last updated on Mar 14, 2023, 12:23 AM UTC (9335f16)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/227/d52ee70...jyasskin:9335f16.html" title="Last updated on Mar 14, 2023, 12:23 AM UTC (9335f16)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1104,7 +1104,12 @@ recognition|recognized=] across [=partitions=] unless they intend to be
 recognized. Note that sites can do harm even if they can't be completely certain
 that visits come from the same person, so [=user agents=] should also take steps
 to prevent such probabilistic recognition. The [[[Privacy-Threat]]] discusses
-the tradeoffs involved.
+the tradeoffs involved. Note that if a [=user agent=] blocks [=recognition=]
+between different sites but doesn't notice when a single site contains multiple
+[=contexts=], this can encourage [=actors=] to bundle multiple contexts into a
+single site and privilege the [=actors=] who can do so. These are both harmful,
+but that harm is likely less than the harm from allowing [=cross-site
+recognition=].
 
 If a [=user agent=] can tell that its user is using a particular identity on a
 website, for example because the user used an API like
--- a/index.html
+++ b/index.html
@@ -1101,15 +1101,20 @@ identities per subdomain or site path.
 
 Where possible, [=user agents=] should prevent people from being [=cross-context
 recognition|recognized=] across [=partitions=] unless they intend to be
-recognized. Note that sites can do harm even if they can't be completely certain
-that visits come from the same person, so [=user agents=] should also take steps
-to prevent such probabilistic recognition. The [[[Privacy-Threat]]] discusses
-the tradeoffs involved. Note that if a [=user agent=] blocks [=recognition=]
-between different sites but doesn't notice when a single site contains multiple
-[=contexts=], this can encourage [=actors=] to bundle multiple contexts into a
-single site and privilege the [=actors=] who can do so. These are both harmful,
-but that harm is likely less than the harm from allowing [=cross-site
-recognition=].
+recognized. Note that:
+
+* Sites can do harm even if they can't be completely certain
+  that visits come from the same person, so [=user agents=] should also take steps
+  to prevent such probabilistic recognition. The [[[Privacy-Threat]]] discusses
+  the tradeoffs involved.
+* If a [=user agent=] blocks [=recognition=] between different sites but doesn't
+  notice when a single site contains multiple [=contexts=], this can encourage
+  [=actors=] to bundle multiple contexts into a single site and privilege the
+  [=actors=] who can do so. Despite these harms, even if a [=user agent=] can't
+  always block [=cross-context recognition=] within a single site, it should
+  still prevent [=cross-site recognition=] when it can. By having [=user
+  agents=] solve the problems that technical measures can solve, we let actors
+  using other approaches focus on the problems they're best at.
 
 If a [=user agent=] can tell that its user is using a particular identity on a
 website, for example because the user used an API like
--- a/index.html
+++ b/index.html
@@ -1095,9 +1095,10 @@ container, or container tab for user agents that support those features)
 cookies and other storage (which is sometimes automatic at the end of each
 session).
 
-When a [=user agent=] knows that a site includes multiple contexts, it
-should adjust its [=partitions=] accordingly, for instance by partitioning
-identities per subdomain or site path.
+When a [=user agent=] knows that a site includes multiple contexts, it should
+adjust its [=partitions=] accordingly, for instance by partitioning identities
+per subdomain or site path. User agents should work to improve their ability to
+distinguish contexts within a site.
 
 Where possible, [=user agents=] should prevent people from being [=cross-context
 recognition|recognized=] across [=partitions=] unless they intend to be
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1104,7 +1104,12 @@
 recognized. Note that sites can do harm even if they can't be completely certain
 that visits come from the same person, so [=user agents=] should also take steps
 to prevent such probabilistic recognition. The [[[Privacy-Threat]]] discusses
-the tradeoffs involved.
+the tradeoffs involved. Note that if a [=user agent=] blocks [=recognition=]
+between different sites but doesn't notice when a single site contains multiple
+[=contexts=], this can encourage [=actors=] to bundle multiple contexts into a
+single site and privilege the [=actors=] who can do so. These are both harmful,
+but that harm is likely less than the harm from allowing [=cross-site
+recognition=].
```
darobin, line 1112: The change is generally okay but I feel like the conclusion is a little weak.

Maybe replace that last sentence with: `The fact that [=user agents=] can't mitigate all issues does not entail that they should mitigate none. On the contrary, mitigating an important part of the problem makes it easier for other approaches (eg. regulatory) to focus on what technical means cannot address.` Feel free to change the words if you like the change, of course, the point is to convey that this is an "and" approach, and that if regulators don't have to worry about cookies they could target more interesting things.

index.html
```diff
@@ -1104,7 +1104,12 @@
 recognized. Note that sites can do harm even if they can't be completely certain
 that visits come from the same person, so [=user agents=] should also take steps
 to prevent such probabilistic recognition. The [[[Privacy-Threat]]] discusses
-the tradeoffs involved.
+the tradeoffs involved. Note that if a [=user agent=] blocks [=recognition=]
+between different sites but doesn't notice when a single site contains multiple
+[=contexts=], this can encourage [=actors=] to bundle multiple contexts into a
+single site and privilege the [=actors=] who can do so. These are both harmful,
+but that harm is likely less than the harm from allowing [=cross-site
```
dmarti, line 1111: Likely more? If user agents block cross-domain recognition, but not cross-context recognition within domains, then single-context domain owners are pressured to either (1) sell out to the owner of a multi-context domain, leading to the formation of large, systemically risky, firms or (2) enter into an agreement to proxy their traffic through a server hosting a multi-context domain, creating an unnecessary single point of failure for multiple web services, along with risk of cross-organization personal info sharing.

Or we can state that both cross-context/same-domain recognition and cross-domain recognition are bad without rating one as worse than the other?

index.html
jyasskin: That seems reasonable. How's my new text? I reformatted into bullets to try to keep the section manageable.

index.html
jyasskin: The change I made for Robin also avoids comparing the harms.

index.html
darobin: LGTM!

darobin (approved):

index.html
dmarti: Thank you--made another comment there (this is tricky, may people will read as being encouraged to consolidate more contexts on fewer domains)

index.html
```diff
@@ -1101,10 +1101,20 @@
...
+* Sites can do harm even if they can't be completely certain
+  that visits come from the same person, so [=user agents=] should also take steps
+  to prevent such probabilistic recognition. The [[[Privacy-Threat]]] discusses
+  the tradeoffs involved.
+* If a [=user agent=] blocks [=recognition=] between different sites but doesn't
+  notice when a single site contains multiple [=contexts=], this can encourage
+  [=actors=] to bundle multiple contexts into a single site and privilege the
+  [=actors=] who can do so. Despite these harms, even if a [=user agent=] can't
+  always block [=cross-context recognition=] within a single site, it should
+  still prevent [=cross-site recognition=] when it can. By having [=user
```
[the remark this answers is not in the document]
jyasskin: That works for me. I split up your wording to put the context-detection part next to our existing text about detecting sub-site contexts.

I am nervous about over-promising here&mdash;I'm not confident browsers will be able to reliably cut up sites into multiple partitions&mdash;but I don't mind saying that we should try.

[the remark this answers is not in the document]
dmarti: Thank you -- I agree, it will be harder for browsers to identify separate contexts within a site than contexts that have their own sites (but blocking fingerprinting is hard too, and browsers can do a lot to deter investments in fingerprinting just by declaring that they're going after it where they can)

torgo (approved):
lgtm

PR #243, opened 2023-03-29
Make it clear that this is a general statement, not about the GDPR

### LINKED ISSUE #232: Don't imply the GPC has specific effects under the GPDR.
@michaelkleber points out that https://w3ctag.github.io/privacy-principles/#opt-in-out has 

> A good example of a [global opt-out](https://w3ctag.github.io/privacy-principles/#dfn-global-opt-out) mechanism is the Global Privacy Control [[GPC](https://w3ctag.github.io/privacy-principles/#bib-gpc)].

followed shortly thereafter by

> under the [[GDPR](https://w3ctag.github.io/privacy-principles/#bib-gdpr)], the [person](https://w3ctag.github.io/privacy-principles/#dfn-individual) may be conveying objections to [processing](https://w3ctag.github.io/privacy-principles/#dfn-process) based on legitimate interest or the withdrawal of [consent](https://w3ctag.github.io/privacy-principles/#dfn-opt-in) to specific [purposes](https://w3ctag.github.io/privacy-principles/#dfn-purpose).

which implies that the GPC has that effect. Since the GPC has never been tested under the GDPR and was designed to match the CCPA, we shouldn't make that legal implication.

darobin:
This makes it clear that this isn't legal advice or a specific position about the GDPR but an understanding of how an automated signal can intersect with specific data protection approaches. Fixes #232.

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/pull/243.html" title="Last updated on Apr 12, 2023, 1:59 PM UTC (d0e96e0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/243/5690bcc...d0e96e0.html" title="Last updated on Apr 12, 2023, 1:59 PM UTC (d0e96e0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -657,10 +657,10 @@ Conceptually, a [=global opt-out=] mechanism is an automaton operating as part o
 [=user agent=], which is to say that it is equivalent to a robot that would carry out a
 [=person=]'s bidding by pressing an [=opt-out=] button with every interaction that the
 [=person=] has with a site, or more generally conveys an expression of the [=person=]'s
-rights in a relevant jurisdiction. (For instance, under the [[?GDPR]], the [=person=] may be
-conveying objections to [=processing=] based on legitimate interest or the withdrawal of
-[=consent=] to specific [=purposes=].) It should be noted that, since a [=global opt-out=]
-signal is reaffirmed automatically with every interaction, it will take precedence
+rights in a relevant jurisdiction. (For instance, under data protection regimes based on legitimate
+interest or consent, the [=person=] may be conveying objections to [=processing=] based on legitimate
+interest or the withdrawal of [=consent=] to specific [=purposes=].) It should be noted that, since a
+[=global opt-out=] signal is reaffirmed automatically with every interaction, it will take precedence
 in terms of specificity over any general obtention of [=consent=] by a site,
 and only superseded by specific [=consent=] obtained through a deliberate action taken by
 the user with the intent of overriding their global opt-out.
--- a/index.html
+++ b/index.html
@@ -657,9 +657,9 @@ Conceptually, a [=global opt-out=] mechanism is an automaton operating as part o
 [=user agent=], which is to say that it is equivalent to a robot that would carry out a
 [=person=]'s bidding by pressing an [=opt-out=] button with every interaction that the
 [=person=] has with a site, or more generally conveys an expression of the [=person=]'s
-rights in a relevant jurisdiction. (For instance, under data protection regimes based on legitimate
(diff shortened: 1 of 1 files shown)


npdoty:
Maybe we don't need a reference to "legitimate interest" but just to note that it could be withdrawing consent or it could be about objecting to some processing, separate from any particular jurisdiction.

index.html
```diff
@@ -657,10 +657,10 @@
 [=user agent=], which is to say that it is equivalent to a robot that would carry out a
 [=person=]'s bidding by pressing an [=opt-out=] button with every interaction that the
 [=person=] has with a site, or more generally conveys an expression of the [=person=]'s
-rights in a relevant jurisdiction. (For instance, under the [[?GDPR]], the [=person=] may be
-conveying objections to [=processing=] based on legitimate interest or the withdrawal of
-[=consent=] to specific [=purposes=].) It should be noted that, since a [=global opt-out=]
-signal is reaffirmed automatically with every interaction, it will take precedence
+rights in a relevant jurisdiction. (For instance, under data protection regimes based on legitimate
+interest or consent, the [=person=] may be conveying objections to [=processing=] based on legitimate
+interest or the withdrawal of [=consent=] to specific [=purposes=].) It should be noted that, since a
```
jyasskin, line 662: If we list things that a global opt-out can do, we should include the thing Sec-GPC [actually does](https://privacycg.github.io/gpc-spec/#the-sec-gpc-header-field-for-http-requests), which is to convey a preference that the person's data not be sold to or shared.
@dmarti's suggestion in https://github.com/w3ctag/privacy-principles/issues/232#issuecomment-1480016252 would also be fine.

dmarti: I like this better than my suggestion -- GDPR, as far as I know, does not mandate a specific format or protocol for an objection to processing, so IMHO it is fine for us to recommend treating GPC as an objection in general.

(This version helps make this document more future-proof -- we don't want it to be too tied to this weird time during which a lot of GDPR "compliance" is still fake, and the cases to close the loopholes haven't worked their way through the process yet)

jyasskin:
```suggestion
rights in a relevant jurisdiction. (For instance, the [=person=] may be objecting to [=processing=]
based on legitimate interest, withdrawing [=consent=] to specific [=purposes=], or requesting that
their data not be sold or shared.) It should be noted that, since a
```

jyasskin (approved):

torgo (approved):
lgtm

PR #249, opened 2023-04-12
Avoid saying that safety systems have to describe how to exploit them.

jyasskin:
The text I originally wrote here got interpreted as saying that services
have to explain all the details of how they use data to fight fraud,
which of course would help the fraudsters avoid detection. Instead, we
meant just that the service should say what kinds of things it's
collecting in order to protect people.

I also got a comment on how abrupt this paragraph is, since it just
restricts the safety-related data collection without saying that this
purpose is reasonable in the first place.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/249.html" title="Last updated on Apr 12, 2023, 4:50 PM UTC (3f4b834)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/249/4b06c6c...jyasskin:3f4b834.html" title="Last updated on Apr 12, 2023, 4:50 PM UTC (3f4b834)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -894,9 +894,7 @@ This is a special case of the more general principle that data should not be use
 [=purposes=] than the data's subjects understood it was being collected for.
 
 Services sometimes use people's data in order to protect those or other people.
-This is legitimate as long as the extra risk created by the data collection is
-less than the protection it offers and the extra risk isn't concentrated on
-[=vulnerable people=]. A service that does this should explain what data it's
+A service that does this should explain what data it's
 using for this purpose. It should also say how it might use or share a person's
 data if it believes that person has broken the rules.
 
--- a/index.html
+++ b/index.html
@@ -893,9 +893,12 @@ for other purposes, like to grow the service.</span></p>
 This is a special case of the more general principle that data should not be used for more
 [=purposes=] than the data's subjects understood it was being collected for.
 
-A service should explain how it uses people's data to protect them and other people, and
-how it might additionally use someone's data if it believes that person has broken the
-rules.
+Services sometimes use people's data in order to protect those or other people.
+This is legitimate as long as the extra risk created by the data collection is
+less than the protection it offers and the extra risk isn't concentrated on
+[=vulnerable people=]. A service that does this should explain what data it's
+using for this purpose. It should also say how it might use or share a person's
+data if it believes that person has broken the rules.
 
 </div>
 


index.html
```diff
@@ -893,9 +893,12 @@
 This is a special case of the more general principle that data should not be used for more
 [=purposes=] than the data's subjects understood it was being collected for.
 
-A service should explain how it uses people's data to protect them and other people, and
-how it might additionally use someone's data if it believes that person has broken the
-rules.
+Services sometimes use people's data in order to protect those or other people.
+This is legitimate as long as the extra risk created by the data collection is
+less than the protection it offers and the extra risk isn't concentrated on
```
darobin, line 898: I agree that the original text is abrupt and your iteration is better. My concern is that the blanket granting of legitimacy is too strong here. Risk isn't a single variable, the trade-offs are often complex. (How do you compare the risk from processing of one extra data point on billions of people versus potential serious harm to one?) Who assesses the risk of data collection? This feels way too open to the classic "it's for your own good" or "it's for the greater good" arguments that we've heard a lot from governments and tech companies.

I don't have an immediate improvement, but is there a way to formulate this that doesn't grant safety teams extrajudicial powers?

jbradleychen: I think a common way to capture the notion of risk vs. reward is the principle of proportionality. So something like "as long as the risk of data retention is measured, proportional to the benefit to users and the public, and the service is accountable for misuse of user data."

index.html
```diff
@@ -893,9 +893,12 @@
...
+[=vulnerable people=]. A service that does this should explain what data it's
```
jbradleychen, line 899: I think this part "should explain" risks being too prescriptive and too narrow. Instead, how about something like "should practice transparency and accountability for responsible use of user data."

jbradleychen, line 899: I'm not a fan of the vulnerable people language. I don't see how a service can support incompatible classes of vulnerable people, such as sex workers and children, with this language. A platform needs the freedom and liberty to make choices about which users they will preference, and that will inevitable disadvantage other users.

jbradleychen (commented):
This is helpful; that you for starting this.

I would encourage the working group to also add something about managing public data for the public interest. I would understand if the working group decided public data was out of scope. That said your efforts are at best philosophical unless services and publishers can be held to account for failures to follow these policies, and accountability generally fails without public data.

To be concrete, something like "A service should make a clear distinction between private data and public data, and provide clear policies about managing public data in the public interest. Such policies are out of scope for this document."

index.html
jyasskin, line 899: I also got encouraged to make this even stronger, citing the [GDPR's](https://www.privacy-regulation.eu/en/recital-47-GDPR.htm) 

> The processing of personal data strictly necessary for the purposes of preventing fraud also constitutes a legitimate interest of the data controller concerned.

and the [CCPA's](https://www.oag.ca.gov/sites/all/files/agweb/pdfs/privacy/oal-sub-final-text-of-regs.pdf)

> A service provider shall not retain, use, or disclose personal information obtained in the
course of providing services except: ... To detect data security incidents or protect against fraudulent or illegal activity;

jyasskin, line 899: In order to merge some improvement quickly, we have a suggestion to:
```suggestion
A service that does this should explain what data it's
```

Then I'll send another PR about the legitimacy of the practice.

darobin (approved):

npdoty:
That was my suggestion that we don't need to say anything about legitimacy here, as the principle already states that sites may do this when they need to. lgtm. Also, I'd appreciate a general principle about purpose limitation.

PR #266, opened 2023-05-04
Tweaks to the Incorporating section

### LINKED ISSUE #256: Feedback on section 1.4 "Incorporating Different Privacy Principles"
The title of this section doesn't really capture what this section describes, which is how to balance or resolve tensions between different principles. I don't have a good suggestion, but "Incorporating" doesn't really capture it .. maybe "Resolving Tensions Between" ?

The W3C TAG Ethical Web Principles should be cited in reference style rather than linked inline (and similar for Target Privacy Threat Model, later in the document).

"if it believes that person has broken the rules" - It's not clear which rules this refers to. Is it the same thing that's described later as "violates the norms of a service they're using"? If so I recommend aligning the language.

An observation on Example 5: This is related to current work in progress ("are trying to...", and discussion of conversion attribution APIs), and so will eventually become dated as things progress. That may be fine if you plan to revisit the guidance in the document in the future, but it could be worth either rephrasing to be less specific to the here and now or moving elsewhere (e.g., input to PATCG).

jyasskin:
- Use [[Citation-Style]] instead of just links.
- Use consistent language around violating norms.

Fixes #256.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/266.html#balancing" title="Last updated on May 10, 2023, 4:03 PM UTC (7575d92)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/266/6b09507...jyasskin:7575d92.html" title="Last updated on May 10, 2023, 4:03 PM UTC (7575d92)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -909,7 +909,7 @@ This is a special case of the more general principle that data should not be use
 Services sometimes use people's data in order to protect those or other people.
 A service that does this should explain what data it's
 using for this purpose. It should also say how it might use or share a person's
-data if it believes that person has violated the service's norms.
+data if it believes that person has violated the service's rules.
 
 </div>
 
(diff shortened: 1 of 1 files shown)


chrisn (approved):

index.html
```diff
@@ -909,7 +909,7 @@
 Services sometimes use people's data in order to protect those or other people.
 A service that does this should explain what data it's
 using for this purpose. It should also say how it might use or share a person's
-data if it believes that person has broken the rules.
+data if it believes that person has violated the service's norms.
```
wseltzer, line 912:
```suggestion
data if it believes that person has violated the service's rules.
```
Adding "the service's" seems useful, but norms looks like a fuzzier bound than rules, providing less certainty to users.

jyasskin: That sounds like a good point to me. To maintain the fix to @chrisn's issue, I changed all the uses of "norms" to "rules" in this section. How does that wind up looking?

chrisn: Seems good to me

darobin (approved):

PR #275, opened 2023-06-27
No excuses for fingerprinting

martinthomson:
The notion that fingerprinting might not be economically feasible doesn't really help anyone.  On the one hand, it just tends to create an excuse for lazy web designers who might choose to neglect fingerprinting protection on the basis that "it won't be viable to exploit this" (tip: it probably is viable).  On the other, the admonition about people who didn't ask to be treated that way is already implied.y

Also, the document has enough words in it already.  No point playing both sides here.

I considered saying something more in place of this, but the paragraph is already complete as is.  That is, it talks about the challenges inherent with fingerprinting in that it disproportionately affects those smaller groupings that tend to form around less mainstream choices or characteristics.

***
<a href="https://pr-preview.s3.amazonaws.com/martinthomson/privacy-principles/pull/275.html" title="Last updated on Jun 27, 2023, 4:47 AM UTC (9e423b3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/275/5286a4e...martinthomson:9e423b3.html" title="Last updated on Jun 27, 2023, 4:47 AM UTC (9e423b3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1188,11 +1188,7 @@ For example:
 Preventing [=fingerprinting=] can be particularly challenging in cases that
 only affect a small group of people who use the web. For example, people who
 configure their systems in unique ways, such as by using a browser with a very
-small number of users. As long as a tracker can't track a significant number
-of people, it's likely to be unviable to maintain the tracker. However, this
-doesn't excuse making small groups of people trackable when those people
-didn't choose to be in the group.
-
+small number of users.
 See [[fingerprinting-guidance]] for how to mitigate threats that result from
 [=fingerprinting=].
 


jyasskin:
Without this bit, I think the document winds up saying that it violates the principle to release a new browser that doesn't have many users yet, or to provide configuration options, even behind `about:config`, that could identify a user.

martinthomson:
No, it just says that avoiding fingerprinting is hard in those cases.  Which is true.

npdoty:
It could be that the separate fingerprinting mitigation guidance doc could go into the details about viability of different kinds of trackers.

Issue #263, opened 2023-05-03
principle and additional context on transparency


npdoty:
Transparency comes up in "Transparency and research" section, and in a few other sections, but could be elaborated and be more specific about what makes transparency relatively effective. That shouldn't contradict the points made about transparency for individuals being insufficient or challenging to rely on.

Transparency about privacy practices should be machine-readable (to facilitate those group or user agent-facilitated decision-making), accessibly presented, contextual and actionable. This could connect to consent/permissions.

PR #288, opened 2023-06-27
transparency section

npdoty:
addresses #263 

principles for providing transparency, plain and machine-readable, distinguishable apis

citations to Adding Permissions guide, Unsanctioned Tracking, Fingerprinting

note with links to permissions workshops


index.html
```diff
@@ -1913,6 +1919,64 @@
 
 </aside>
 
+## Transparency
+
+<div class="practice">
+  <span class="practicelab" id="transparency-when-requested">
+    [=People=] should be provided relevant information when data is accessed or when permission is
+    requested.
```
darobin, line 1927: Might it make sense to  flip this around to be more active? Maybe: "Sites and implementations should provide people with relevant information (about the intended processing) when accessing their data or requesting their permission."

index.html
```diff
@@ -1913,6 +1919,64 @@
...
+  </span>
+</div>
+
+Transparency is a necessary, but insufficient, condition for [=consent=]. Relevant information
+includes who may be accessing information, what information may be accessed (including the potential
+inferences or combinations of such data) and how information may be used. For transparency to be
+meaningful to people, it must be contextual.
```
darobin, line 1934: I worry about all the "mays." I know that this is a legalistic approach, but it tends to lead to bigger bags than necessary and participates in making people believe that a lot of their data needs to be accessed. I'm not sure how to capture that, but I think that transparency would be helped if it were more definitive.

index.html
```diff
@@ -1913,6 +1919,64 @@
...
+  needed and how to make that permission meaningful [[?ADDING-PERMISSIONS]].
+
+  Past workshops have explored the needs for better permissions on the Web:
+  <ul>
+    <li><a href="https://www.w3.org/Privacy/permissions-ws-2022/report">2022 W3C Workshop on
+    Permissions</a></li>
+    <li><a href="https://www.w3.org/Privacy/permissions-ws-2018/report.html">2018 W3C Workshop on
+    Permissions and User Consent</a></li>
+    <li><a href="https://www.w3.org/2014/07/permissions/minutes.html">2014 Next steps on trust and
+    permissions for Web applications</a></li>
```
darobin, line 1947: The 2006 chartering of Web APIs and the 2009 chartering of DAP also had this in scope, if you want to go there :)

index.html
```diff
@@ -1913,6 +1919,64 @@
...
+  </ul>
+</div>
+
+<div class="practice">
+  <span class="practicelab" id="transparency-plain-language-machine-readable">
+    Information about privacy-relevant practices should be provided in both easily accessible plain
+    language form and in machine-readable form.
```
darobin, line 1954: Provided by whom?

index.html
```diff
@@ -1913,6 +1919,64 @@
...
+    language form and in machine-readable form.  
+  </span>
+</div>
+
+Machine-readable presentation of privacy-relevant practices is necessary for [=user agents=] to be
+able to help [=people=] make general decisions, rather than relying falsely on the idea that
+[=people=] can or want to read documentation before every visit to a web site. Machine-readable
+presentation also facilitates <a href="#collective">collective governance</a> by making it more
+feasible for researchers and regulators to discover, document, and analyze data collection and
+processing in ways that may be harmful.
```
darobin, line 1963: I don't think we want to support researchers analysing data in ways that may be harmful :-D

```suggestion
processing to identify cases in which it may be harmful.
```

index.html
jyasskin, line 1927: This feels vague. Which people? What information is relevant? (That's covered below, but it'd be good if the principle could be read alone.) Should we interrupt people whenever a backend system touches their data?

I think we have a couple goals in this section:

1. People should be made aware of (or be able to find out?) what actors have their data and how those actors use it.
2. The above information should be in a form that allows people to automate their reactions to it.
3. Researchers and regulators should be able to aggregate that information (in privacy-preserving ways) in order to analyze how data is used and abused across society.

Maybe the first one would be covered by:

```suggestion
    [=People=] should be able to find out which [=actors=] have their data and how they're using it.
```

On the permissions point, maybe there's something more general about UAs needing to inform users when they give out information about the user, which pushes for designing web APIs that avoid passive fingerprintability and that provide an opportunity to ask the user's permission before disclosing the information.

index.html
jyasskin, line 1934: What does it mean for transparency to be contextual?

index.html
```diff
@@ -1913,6 +1919,64 @@
...
+
+Easily accessible, plain language presentation of privacy-relevant practices is necessary for
+[=people=] to be able to make informed decisions in specific cases when they choose to do so.
+
+<div class="practice">
+  <span class="practicelab" id="transparency-distinguishable">
+    Mechanisms that can be used for tracking or identifying [=people=] should be designed so that
+    their operation is visible and distinguishable, to [=user agents=], researchers and regulators.
```
jyasskin, line 1971: I think this is a special case of the principles above that data use should be machine-readable? We should probably say that. Should this mention [=recognizing=] people?

index.html
```diff
@@ -1913,6 +1919,64 @@
...
+  </span>
+</div>
+
+Unsanctioned tracking is harmful in part because it is not visible to the user, which undermines
+user control [[?UNSANCTIONED-TRACKING]]. Designing features that minimize data and make requests for
```
jyasskin, line 1976: It was odd to see "unsanctioned" here without it being in the principle above. Is there a way to lead into it more smoothly? Or to just say more directly that non-transparent tracking undermines user control and is presumed to be against users' interests since otherwise the tracker could be transparent about it?

index.html
npdoty: This text is broad (perhaps still too broad) in part because the intent is to go beyond discoverability about who has data about you, though that is good to call out separately as well.

Sites need to provide, and UAs may need to help present or consume, relevant information about data collection and use when collection is happening. Permissions is a special case of that where the APIs are easier to reason about, but sites also collect data in the course of navigation and in using HTML functionality (forms, etc.).

npdoty:
In a few cases here, the text uses passive voice, which was helpful in that it was broad, but reviewers point out that its vague.

I think what we typically mean is: sites should provide, and user agents should help present and consume, information to people

But there are some cases where it might go beyond that typical case. A user agent might collect or share information and should itself provide transparency. Other actors may collect or share information outside of the site-browser interaction. Should we exclude these from the scope of the document? Or just talk about the typical case at the top, and then add a sentence or two about other cases where different actors need to cooperate in providing transparency?

npdoty:
Maybe "actors who are processing data (typically, sites) should provide, and user agents should help present and consume,"

index.html
npdoty: Made it clear that it's the information that would be contextual, and instead described it as presenting the information in the relevant context (since context is a defined term). But this could still be clarified. And it may not be necessary for this document, even if it seems like generally good advice.

index.html
```diff
@@ -1913,6 +1919,67 @@
 
 </aside>
 
+## Transparency
+
+<div class="practice">
+  <span class="practicelab" id="transparency-when-requested">
+    When accessing data or requesting permission, [=sites=] (and other [=actors=]) should provide
+    [=people=] with relevant information about its use and [=user agents=] should help present and
```
jyasskin, line 1927: I'm not sure exactly how to best fix this, but I stumbled over the pairing of "its use" with both "accessing data" and "requesting permission". We could make it "requesting permission to access data". Or "information about the site's plans" if the permission might also be about things like storing data. Or I'm not sure that "requesting permission" belongs here, since transparency is about what the sites do with the thing they got permission for.

index.html
```diff
@@ -1913,6 +1919,67 @@
...
+    consume that information.
+  </span>
+</div>
+
+Transparency is a necessary, but insufficient, condition for [=consent=]. Relevant information
```
jyasskin, line 1932: The first two sentences in this paragraph don't seem to have much to do with each other.

index.html
```diff
@@ -1913,6 +1919,67 @@
...
+includes who is accessing information, what information is accessed (including the potential
```
jyasskin, line 1933: Here we have two different meanings of "information" in the same sentence. First the information/explanation that needs to be given to a user, and second the information whose use the user is being informed about.

Would it help to call this a "helpful explanation" instead of "relevant information"?

index.html
```diff
@@ -1913,6 +1919,67 @@
...
+inferences or combinations of such data) and how information is used. For transparency to be
+meaningful to people, information must be provided in the relevant [=context=].
```
jyasskin, line 1935: It's not completely clear whether the "information" that must be provided in context is 1) the information that gets accessed or 2) the "relevant information" that helps users make informed choices. Contexts are used in the rest of the document to provide principles about information use, which leans toward the first interpretation, but I think you mean the second.

index.html
```diff
@@ -1913,6 +1919,67 @@
 
 </aside>
 
+## Transparency
+
+<div class="practice">
+  <span class="practicelab" id="transparency-when-requested">
+    When accessing data or requesting permission, [=sites=] (and other [=actors=]) should provide
+    [=people=] with relevant information about its use and [=user agents=] should help present and
+    consume that information.
```
jyasskin, line 1928: Browsers have historically shied away from showing a site's explanation of its permission request alongside the permission request, mostly because we can't guarantee that the site is telling the truth and don't want to endorse their claim. I'm worried that this principle is going to argue against that practice.

index.html
```diff
@@ -1913,6 +1919,67 @@
...
+  <span class="practicelab" id="transparency-distinguishable">
+    Mechanisms that can be used for [=recognize|recognizing=] [=people=] should be designed so that
+    their operation is visible and distinguishable, to [=user agents=], researchers and regulators.
+  </span>
+</div>
+
+Non-transparent methods of [=recognition=] are harmful in part because they are not visible to the
+user, which undermines user control [[?UNSANCTIONED-TRACKING]]. Designing features that minimize
+data and make requests for data explicit can enable detectability, a kind of transparency that is an
+important mitigation for <a>browser fingerprinting</a>.
```
jyasskin, line 1981: Making the voice more active:

```suggestion
user, which undermines user control [[?UNSANCTIONED-TRACKING]]. When web features are designed
so that sites explicitly request the [[#data-minimization|minimal]] amount of data that they
need, they make sites' data use "detectable", a kind of transparency that helps
mitigate <a>browser fingerprinting</a>.
```

index.html
npdoty: Agree that this was ambiguous over "information". Thinking about phrasing, but "relevant explanation", "explanatory text", something like that.

index.html
npdoty: No strong preference, but the suggestion here seems to convert to passive voice, not active.

index.html
npdoty: Help in presenting and consuming information doesn't require any particular UI about permissions requests. I think we're agreed that this wouldn't prohibit any current browser vendor preference about UI.

(Although, also, yes, browsers should change their practices to improve visibility of the site's explanation when permissions are being requested, as the current model of providing no context or explanation whatsoever has proved extremely ineffective.)

index.html
npdoty: Tried to clarify this in recent commit, "with relevant explanatory information about the use of data".

index.html
npdoty: Relevant explanatory information is an example of providing the transparency necessary for consent.

index.html
npdoty: Clarified with "explanatory information".

index.html
```diff
@@ -1913,6 +1919,67 @@
...
+
+Easily accessible, plain language presentation of privacy-relevant practices is necessary for
+[=people=] to be able to make informed decisions in specific cases when they choose to do so.
+[=Sites=], [=user agents=], and other [=actors=] all may need to present privacy-relevant practices
+to [=people=] in accessible forms.
+
+<div class="practice">
+  <span class="practicelab" id="transparency-distinguishable">
+    Mechanisms that can be used for [=recognize|recognizing=] [=people=] should be designed so that
+    their operation is visible and distinguishable, to [=user agents=], researchers and regulators.
```
jyasskin, line 1974:
```suggestion
    their operation is visible and distinguishable to [=user agents=], researchers, and regulators.
```

jyasskin:
We'll merge this once @pes10k and @jyasskin have re-read the text and approved the PR.

index.html
```diff
@@ -1913,6 +1919,67 @@
...
+  </span>
+</div>
+
+Non-transparent methods of [=recognition=] are harmful in part because they are not visible to the
+user, which undermines user control [[?UNSANCTIONED-TRACKING]]. Designing features that minimize
+data and make requests for data explicit can enable detectability, a kind of transparency that is an
```
pes10k, line 1980: i dont think its accurate that detectability is a mitigation for browser fingerprinting.  Its often easy to detect when scripts are fingerprinting a user; but that doesn't mitigate / prevent the fingerprinting from happening

pes10k (approved):
I left one substantive comment, but LGTM, and happy to address that in follow up too

index.html
npdoty: The [Mitigating Browser Fingerprinting](https://www.w3.org/TR/fingerprinting-guidance/#detectability) doc describes detectability as one mitigation, though clearly not a solution or prevention of fingerprinting altogether. If we want to come up with another word to describe something that can be a useful response to a problem without solving it, I'm open to that, but if so we should (also) file an issue on the fingerprinting-guidance.

jyasskin (approved):
We could still improve this section's readability, but that doesn't need to block getting the concepts in.

PR #291, opened 2023-06-28
change wording around self-governing

### LINKED ISSUE #233: Clarify uses of "self-governing"
In a comment, @michaelkleber understood "self-governing" to refer to approaches where an industry determines the rules for its own behavior, and I think that matches broad usage in industry. In https://w3ctag.github.io/privacy-principles/#privacy-labour, I think we instead mean approaches where users are expected to do their own work to govern how data about them leaks out into the world and is used.

We should be clear about what meaning we intend, and I suspect the best way to do that is to spell it out instead of using "self-governing" at all.

torgo:
This is pretty minimal. It just re-words the sentence to include the definition first before we mention the term self-governing. Fixes #233.

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/pull/291.html" title="Last updated on Jun 28, 2023, 2:25 PM UTC (dc412f9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/291/86d1c1f...dc412f9.html" title="Last updated on Jun 28, 2023, 2:25 PM UTC (dc412f9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -691,8 +691,8 @@ the work of ensuring [=data processing=] of which they are the subject or recipi
 Data systems that are based on asking [=people=] for their [=consent=] tend to increase
 [=privacy labour=].
 
-More generally, implementations of [=privacy=] are often dominated by self-governing approaches that
-offload [=labour=] to [=people=]. This is notably true of the regimes descended from the
+More generally, implementations of [=privacy=] are often dominated by approaches that
+offload [=labour=] to [=people=] (self-governing). This is notably true of the regimes descended from the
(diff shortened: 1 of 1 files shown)


michaelkleber:
Ah I see!  I definitely misunderstood what this meant, thanks for the edit.

Where is "self-governing" used like this?  I've heard it called "privacy self-management", e.g. by Solovay in https://papers.ssrn.com/sol3/papers.cfm?abstract_id=2171018.  (Is there a canonical reference for "self-governing" that we can cite?)

darobin:
Overtaken by later edit that gets rid of the word entirely.

PR #238, opened 2023-03-23
Remove the statement about a general opt-out overriding a specific opt-in.

jyasskin:
Fixes #231, but I could also add a note that sites shouldn't ask users to opt back in without a good reason.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/238.html" title="Last updated on Apr 12, 2023, 5:11 PM UTC (a85f896)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/238/be76d52...jyasskin:a85f896.html" title="Last updated on Apr 12, 2023, 5:11 PM UTC (a85f896)">Diff</a>

PR #296, opened 2023-06-28
Revisit global opt out, fix #231

### LINKED ISSUE #231: Don't claim general actions override specific ones
@michaelkleber points out that when https://w3ctag.github.io/privacy-principles/#opt-in-out says

> since a [global opt-out](https://w3ctag.github.io/privacy-principles/#dfn-global-opt-out) signal is reaffirmed automatically with every interaction, it will take precedence in terms of specificity over any general obtention of [consent](https://w3ctag.github.io/privacy-principles/#dfn-opt-in) by a site, and only superseded by specific [consent](https://w3ctag.github.io/privacy-principles/#dfn-opt-in) obtained through a deliberate action taken by the user with the intent of overriding their global opt-out.

it's confusing the fact that the technology sends a signal repeatedly, with the fact that the user only flipped a switch one time to cause it to do so. If a user makes a general setting in their browser, and a contrary setting on a specific website, it's the specificity of the *user's* actions that matter, not the number of times the browser repeats those actions, which means the website-specific setting overrides the global browser setting.

https://privacycg.github.io/gpc-spec/#legal-effects says that sites should disclose how they treat "conflicts between the signal and other specific privacy choices that the person has already made directly with the publisher", which avoids this inversion.

darobin:
This is intended to supersede #238.

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/pull/296.html" title="Last updated on Jun 28, 2023, 6:02 PM UTC (c724a30)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/296/86d1c1f...c724a30.html" title="Last updated on Jun 28, 2023, 6:02 PM UTC (c724a30)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -681,7 +681,7 @@ their data not be sold or shared.) The [=user=] is effectively delegating the ex
 of their [=opt-out=] to their [=user agent=], which helps rectify [=automation asymmetry=].
 
 Under this model, a [=global opt-out=] signal should not be understood as a decision that a
-[=person=] made a while ago when they flipped a setting or chose to user a specific
+[=person=] made a while ago when they flipped a setting or chose to use a specific
 [=user agent=] but rather as a decision that they have chosen to automatically reaffirm with
 every interaction.
 
--- a/index.html
+++ b/index.html
@@ -682,7 +682,7 @@ of their [=opt-out=] to their [=user agent=], which helps rectify [=automation a
 
 Under this model, a [=global opt-out=] signal should not be understood as a decision that a
 [=person=] made a while ago when they flipped a setting or chose to use a specific
-[=user agent=] but rather as a decision that they have chosen to automatically reaffirm with
+[=user agent=] but rather as a preference that they have chosen to automatically reaffirm with
 every interaction.
 
 ### Privacy Labour {#privacy-labour}
--- a/index.html
+++ b/index.html
@@ -677,11 +677,13 @@ Conceptually, a [=global opt-out=] mechanism is an automaton operating as part o
 [=person=] has with a site, or more generally conveys an expression of the [=person=]'s
 rights in a relevant jurisdiction. (For instance, the [=person=] may be objecting to [=processing=]
 based on legitimate interest, withdrawing [=consent=] to specific [=purposes=], or requesting that
-their data not be sold or shared.) It should be noted that, since a
-[=global opt-out=] signal is reaffirmed automatically with every interaction, it will take precedence
-in terms of specificity over any general obtention of [=consent=] by a site,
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -677,11 +677,13 @@
...
-[=global opt-out=] signal is reaffirmed automatically with every interaction, it will take precedence
-in terms of specificity over any general obtention of [=consent=] by a site,
-and only superseded by specific [=consent=] obtained through a deliberate action taken by
-the user with the intent of overriding their global opt-out.
+their data not be sold or shared.) The [=user=] is effectively delegating the expression
+of their [=opt-out=] to their [=user agent=], which helps rectify [=automation asymmetry=].
+
+Under this model, a [=global opt-out=] signal should not be understood as a decision that a
+[=person=] made a while ago when they flipped a setting or chose to user a specific
+[=user agent=] but rather as a decision that they have chosen to automatically reaffirm with
```
darobin, line 685:
```suggestion
[=user agent=] but rather as a preference that they have chosen to automatically reaffirm with
```

index.html
```diff
@@ -677,11 +677,13 @@
...
-their data not be sold or shared.) It should be noted that, since a
-[=global opt-out=] signal is reaffirmed automatically with every interaction, it will take precedence
-in terms of specificity over any general obtention of [=consent=] by a site,
-and only superseded by specific [=consent=] obtained through a deliberate action taken by
-the user with the intent of overriding their global opt-out.
+their data not be sold or shared.) The [=user=] is effectively delegating the expression
+of their [=opt-out=] to their [=user agent=], which helps rectify [=automation asymmetry=].
+
+Under this model, a [=global opt-out=] signal should not be understood as a decision that a
+[=person=] made a while ago when they flipped a setting or chose to user a specific
```
michaelkleber, line 684:
```suggestion
[=person=] made a while ago when they flipped a setting or chose to use a specific
```

michaelkleber:
Thanks, I'm glad that this is now focused on the user's intent rather than irrelevant implementation details.  And I support being silent on the question of what supersedes what.

As always, I'm nervous about expansive claims about what the user intended, in a context where every user agent and/or other global privacy control provider can present a different UI preference experience (or even none at all).   But delegating this to the user agent is better than any other approach I can think of.

darobin:
Right — that's why I tried to frame it in terms of the conceptual model of what a GOO is. The user is and will remain, ultimately, unknowable but I think that capturing the conceptual model of a GOO can help UAs make the right decisions in terms of what to present where.

jyasskin (approved):

PR #321, opened 2023-07-02
Preview: all copy edits together

rhiaro:
A PR with all of the copy edits across all sections (so far). Individual PRs for each section can be reviewed and merged separately. This is just for easier previewing as a whole.

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/pull/321.html" title="Last updated on Jul 5, 2023, 5:03 PM UTC (bf18054)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/321/21b633a...bf18054.html" title="Last updated on Jul 5, 2023, 5:03 PM UTC (bf18054)">Diff</a>

PR #323, opened 2023-07-02
302 edit identity

rhiaro:
Moves Recognition definition and subsections into its own section in the Common Concepts appendix.

See https://github.com/w3ctag/privacy-principles/pull/321 for edits in context.

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/pull/323.html" title="Last updated on Jul 12, 2023, 4:11 PM UTC (3930eca)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/323/077d2a9...3930eca.html" title="Last updated on Jul 12, 2023, 4:11 PM UTC (3930eca)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1041,194 +1041,36 @@ if the same data were shared between unrelated [=actors=].
 
 <span class="practicelab" id="principle-identity-per-context">A [=user agent=]
 should help its user present the [=identity=] they want in each [=context=]
-they are in.</span>
+they are in, and to prevent or support [=recognition=] as appropriate.</span>
 
 </div>
 
 A [=person=]'s <dfn>identity</dfn> is the set of characteristics that define
 them. Their identity *in a [=context=]* is the set of characteristics they
-present in that context. People frequently present different identities to
-different contexts, and also frequently share an identity among several
-contexts. People may also wish to present an ephemeral or anonymous identity,
-which is just a set of characteristics that is too small or unstable to be useful
+present under particular circumstances.
+
+People can present different identities to different contexts, and can
+also use a single identity in several different contexts.
+
+People may wish to present an ephemeral or anonymous identity. This is
+a set of characteristics that is too small or unstable to be useful
 for following them through time.
 
-It is important to keep in mind that a person's [=identities=] may often be distinct
-from whatever legal identity or identities they may hold.
+A person's [=identities=] may often be distinct from whatever legal identity
+or identities they may hold.
 
-<dfn data-lt="recognize|recognized">Recognition</dfn> is the act of realising that a given [=identity=]
-corresponds to the same [=person=] as another [=identity=] which may have been
-observed either in another [=context=] or in the same [=context=] but at a
-different time. A [=person=] can be [=recognized=] whether or not their legal identity or
-characteristics of their legal identity are included in the recognition.
+In some circumstances, the best way for a [=user agent=] to uphold this
+principle is to prevent [=recognition=] (i.e. so that one [=site=] can't
+learn anything about its [=user=]'s behavior on *another* site).
 
-In order to uphold the above <a
-href="#principle-identity-per-context">principle</a>, sometimes a [=user agent=]
-needs to *prevent* [=recognition=], for instance so that one [=site=] can't
-learn anything about its [=user=]'s behavior on *another* site. Other times, the
-[=user agent=] needs to *support* [=recognition=],
-for instance to help its [=user=] *prove* to one [=site=] that they have a
-particular identity on another [=site=]. Similarly, a [=user agent=] can help its
-[=user=] to separate or communicate [=identity=] across *repeat* visits to the *same*
-[=site=].
+In other circumstances, the best way for a [=user agent=] to uphold this
+principle is to *support* [=recognition=] (i.e. to help its [=user=] prove
+to one [=site=] that they have a particular identity on another [=site=]).
 
-### Recognition Types {#recognition-types}
+Similarly, a [=user agent=] can help its [=user=] by preventing or supporting
+[=recognition=] across *repeat* visits to the *same* [=site=].
 
-There are several types of [=recognition=] that may take place. These rely on
-different methods and present different challenges.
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2247,6 +2091,175 @@
   monitoring service integrity, load balancing, capacity planning, or billing); and,
 * has a contract in place with the [=actor=] which is consistent with the above limitations.
 
+## Recognition {#recognition}
+
+<dfn data-lt="recognize|recognized">Recognition</dfn> is the act of realising that a given [=identity=]
+corresponds to the same [=person=] as another [=identity=] which may have been
+observed either in another [=context=], or in the same [=context=] but at a
+different time.
```
jyasskin, line 2099: Down by "Cross-site recognition", we have

> when a [=site=] determines with high probability that a visit to the site comes from the same person as another visit to a *different* site.

The high probability bit probably belongs with recognition in general, rather than just with cross-site recognition... Maybe:

```suggestion
<dfn data-lt="recognize|recognized">Recognition</dfn> is the act of realising that a given [=identity=]
corresponds to the same [=person=] as another [=identity=] which may have been
observed either in another [=context=], or in the same [=context=] but at a
different time. [=Recognition=] can be probabilistic, if someone realises there's
a high probability that two [=identities=] correspond to the same [=person=],
even if they aren't certain.
```

index.html
```diff
@@ -2247,6 +2091,175 @@
...
+violate the principles around use of information acquired in a different
+[=context=]. This is particularly true for [=vulnerable=] people, as
+recognising them in different [=contexts=] may force traits into the open
+that reveal their vulnerability. For example, if you meet your therapist at a
+party, you expect them to have different discussion topics with you than they
+usually would, and possibly even to pretend they don't know you.
+
+<dfn>Cross-site recognition</dfn> is when a [=site=] determines with high
+probability that a visit to the site comes from the same person as another
+visit to a *different* site. In the usual case that the sites are different [=contexts=],
```
jyasskin, line 2134: If we move the high-probability bit up to the main definition of recognition, this becomes:

```suggestion
<dfn>Cross-site recognition</dfn> is [=recognition=] when the identities
are observed on different [=sites=]. In the usual case that the sites are
different [=contexts=],
```

index.html
```diff
@@ -2247,6 +2091,175 @@
...
+[=cross-site recognition=] is a privacy harm in the same cases as [=cross-context recognition=].
+
+<dfn>Same-site recognition</dfn> is when a single [=site=] discovers and uses the
+fact that two or more visits probably came from the same [=person=].
```
jyasskin, line 2138:
```suggestion
<dfn>Same-site recognition</dfn> is when a single [=site=] [=recognizes=] a
[=person=] across two or more visits.
```

index.html
```diff
@@ -2247,6 +2091,175 @@
...
+session).
+
+It can be difficult for a [=user agent=] to detect when a single site contains
+multiple [=contexts=]. When a [=user agent=] can detect this, it should adjust
+its [=partitions=] accordingly, for instance by partitioning identities
+per subdomain or site path. User agents should work to improve their ability to
+distinguish contexts within a site.
+
+[=User agents=] should prevent people from being [=cross-context
+recognition|recognized=] across [=partitions=] unless they intend to be
```
jyasskin, line 2177: I think the `data-lt` above lets us:

```suggestion
[=User agents=] should prevent people from being
[=recognized=] across [=partitions=] unless they intend to be
```

index.html
```diff
@@ -2247,6 +2091,175 @@
...
+{{CacheStorage}}, and other forms of storage.
+
+Sometimes sites use this to save the [=person=]'s preferences or ongoing activities
+(e.g. shopping carts), and people have come to expect this behavior in some
+[=contexts=].
+
+People are unlikely to expect [=recognition=] all the time, and can find it
+difficult to prevent or mitigate, especially if it is automated.
+
+[=Recognition=] can be automated in different ways:
```
jyasskin, line 2203: Automation isn't quite the issue here, since first-party cookies are also automated. Maybe it's just when recognition uses ["unsanctioned"](https://www.w3.org/2001/tag/doc/unsanctioned-tracking/) methods, which we've extended to mean 3p cookies, link decoration, etc? I don't have a great idea for rewording, and you didn't add this concept, so I'm just noting the issue. We don't have to fix it to merge this PR.

jyasskin (commented):
Thanks! I'm not sure that recognition is really a "common concept", so I might support leaving it in a subsection of Identity, but the edits themselves look like a big improvement.

index.html
```diff
@@ -1041,194 +1041,38 @@
...
+[=recognition=] across *repeat* visits to the *same* [=site=].
 
-<dfn>Cross-device communication</dfn> is communication
-between code on one device and code running on another device. For example, sounds or
-light emitted from one device could be detected by a microphone or light sensor on
-another device [[?SILVERPUSH]]. [=Cross-device communication=] enables cross-device
-tracking, a form of [=cross-context recognition=], but it can also be used for other
-inappropriate information flows.
+[=User agents=] should do their best to distinguish [=contexts=] within a site,
+and adjust their [=partitions=] when necessary.
```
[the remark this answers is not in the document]
rhiaro: On the fence about this. I haven't been assuming people will read the document from top to bottom in order (who does that??) and partitions is linked to a definition for anyone who hasn't read that bit yet or doesn't know the concept. I am hoping we can avoid re-explaining everything everywhere, as if each section needs to completely stand on its own.

That said, happy for the group to resolve this suggestion without me today!

rhiaro:
One thing left to resolve, all other suggestions accepted

index.html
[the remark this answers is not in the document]
jyasskin:
```suggestion
[=User agents=] should do their best to distinguish [=contexts=] within a site
and adjust their [=partitions=] to prevent or support [=recognition=] across those intra-site [=contexts=]
according to their [=users=]' wishes.
```

darobin (approved):

PR #325, opened 2023-07-02
Editorial: Information Access, fixes #304

### LINKED ISSUE #304: Copy edit section 2.3, Information Access

rhiaro:
Condenses section into a single principle, with explanations. Removes many words.

See https://github.com/w3ctag/privacy-principles/pull/321 for edits in context.

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/pull/325.html" title="Last updated on Sep 6, 2023, 4:29 PM UTC (fed1cbb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/325/077d2a9...fed1cbb.html" title="Last updated on Sep 6, 2023, 4:29 PM UTC (fed1cbb)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1404,8 +1404,8 @@ better not to expose.
 </aside>
 
 New APIs that unavoidably provide access to this kind of information should
-not make that information easier to access, and nor should these APIs be
-easier to use by developers, compared to existing comparable web platform features.
+not make that information easier to access compared to existing comparable
+web platform features.
 
 These APIs should also:
 
--- a/index.html
+++ b/index.html
@@ -1407,7 +1407,7 @@ New APIs that unavoidably provide access to this kind of information should
 not make that information easier to access compared to existing comparable
 web platform features.
 
-These APIs should also:
+Specifications describing these APIs should also:
 
 * make it clear how to remove this access in the event that future web
   platform changes make it possible to remove other access to the same
--- a/index.html
+++ b/index.html
@@ -1359,6 +1359,10 @@ to ensure the API exposes information in [=appropriate=] ways.
   For example, the URLs of resources, the timing of link clicks, and the referrer chain within a
   single origin are not guarded by anything; the scroll position is guarded by the setting to turn off
   JavaScript; and access to the camera or geolocation are guarded by permission prompts.
+  
+  When the `<img loading=lazy>` attribute was added, the designers realized
+  that it exposed the scroll position, so it's also guarded by the setting to turn
+  off JavaScript.
 </aside>
 
 New APIs which add new ways of getting information must be
--- a/index.html
+++ b/index.html
@@ -1343,115 +1343,92 @@ browser fingerprint.
 
 ## Information access {#information}
 
-The many APIs available to websites expose lots of data that can be combined into
-information about people, web servers, and other things. We can divide that
-information into three categories:
-
-1. Information that's fine to expose, for example because a person or group with
-    sufficient authority intended to expose that information or to do something that
-    necessarily exposes the information, or because it's not about people at all. For
-    example:
-
-    * The geolocation and camera APIs ask whether a person wants to expose their data.
-    * The URL a person is visiting must be sent to a server in order to navigate to that
-       URL, and known private-information-retrieval methods are too expensive to avoid that
-       exposure.
-    * The distribution of [[[largest-contentful-paint]]] timings for a website is about a website
-       rather than about the people browsing it, even if the data that informs that
-       measure can also reveal information about people.
-
-1. Information that we don't want to expose and have a plausible plan for removing access
-    to. For example, browsers are gradually removing the ability to join identities
-    between different [=partitions=].
-
-1. Information that we'd rather not expose, but that we don't have a plausible plan for
-    removing access to. For example:
-
-    * Some users are disappointed that the page they're visiting can discover which link
-        they clicked to leave that page. We can't block that information
-        because the page can use <a data-cite="RFC9110#status.3xx">HTTP redirects</a> to
-        learn it, and redirection is a core feature of the web.
-    * Some users are disappointed that a page with permission to run JavaScript can record
-        their pattern of interaction with that page. However, the page does this by using
-        the same events it would use to make the page interactive, so we can't block this
-        information access either.
-
-Some functionality of the web has historically been provided using functional primitives (e.g.
-third-party cookies) that can undermine people's privacy. As explained in <em>Improving the web
-without third-party cookies</em> ([[?web-without-3p-cookies]]), "<em>It is better to approach [these
-use cases] with replacement technologies that are designed-for-purpose and built to respect user
-privacy.</em>"
-
-The following subsections discuss how to review an API proposal that exposes data that
-provides a new way to infer each of the above categories of information. They explain how
-to <a data-cite="design-principles#leave-the-web-better">leave the web better than you
-found it</a>.
-
-### Handling acceptable information {#acceptable-information}
-
 <div class="practice">
-  <span class="practicelab">New APIs can add ways of getting acceptable
-  information that are guarded at least as strongly as the existing ways.</span>
+  <span class="practicelab">Web APIs must not weaken [=access guards=] which
+  protect users from [=inappropriate=] exposure of their information.</span>
 </div>
 
-Acceptable information exposure is always qualified by the (possibly empty) set of
-user-controlled settings or permissions that <dfn data-lt="access guard">guard</dfn> access to it.
-For example, the URLs of resources, the timing of link clicks, and the referrer chain within a
-single origin are not guarded by anything; the scroll position is guarded by the setting to turn off
-JavaScript; and access to the camera or geolocation are guarded by permission prompts.
+The many APIs available to websites expose lots of data that can be combined
+into information about people, web servers, and other things.
+
+User-controlled settings or permissions can <dfn data-lt="access guard">guard
+access</dfn> to data on the web. Whether or not exposing information is
+[=appropriate=] depends on the suitability of these [=access guards=] in
+[=context=].
+
+<aside class="example">
+  For example, the URLs of resources, the timing of link clicks, and the referrer chain within a
+  single origin are not guarded by anything; the scroll position is guarded by the setting to turn off
+  JavaScript; and access to the camera or geolocation are guarded by permission prompts.
+</aside>
+
+New APIs which add new ways of getting information must be
+[=access guards|guarded=] at least as strongly as the existing ways.
 
 Information that would be acceptable to expose under one set of [=access guards=] might be
-unacceptable under another set, so when an API designer intends to explain that their new
+unacceptable under another set. When an API designer intends to explain that their new
 API is acceptable because an existing acceptable API already exposes the same information,
 they must be careful to ensure that their new API is only available under a set of guards
-that's at least as strict. Without those guards, they need to make the argument from
+that are at least as strict. Without those guards, they need to make the argument from
 scratch, without relying on the existing API.
 
-### Handling information that's being removed {#unacceptable-information}
+If existing APIs provide access to some information, but there is a plan to
+change those APIs to prevent that access, new APIs must not be added that
+provide that same information, unless it is possible to put additional
+[=appropriate=] [=access guards=] in place.
 
-<div class="practice">
-  <span class="practicelab">If existing APIs provide access to some information, but
-  we have a plan to change those APIs to prevent that access, new APIs must not be added
-  that provide that same information without extra [=access guards=] that make the access to the
-  information acceptable.</span>
-</div>
+For example, browsers are gradually removing the ability to join identities
+between different [=partitions=]. It is important that new APIs do not add
+features which re-enable [=cross-context recognition=].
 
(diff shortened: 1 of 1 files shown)


rhiaro:
@jyasskin I realise there might be a specific browser-engine-y reason this was three distinct principles, so apologies if I've lost something important by trying to turn them into one! It did read a bit like it was trying to address something/someone very specific, and in doing so obfuscated the "key message", which is what I've tried to re-surface with the edits.

index.html
```diff
@@ -1343,115 +1343,92 @@
...
+
+  Some users are disappointed that a page with permission to run JavaScript can record
+  their pattern of interaction with that page. However, the page does this by using
+  the same events it would use to make the page interactive, so we can't block this
+  information access either.
+</aside>
+
+New APIs that unavoidably provide access to this kind of information should
+not make that information easier to access, and nor should these APIs be
+easier to use by developers, compared to existing comparable web platform features.
```
jyasskin, line 1403: It's not quite that the APIs shouldn't be easier to use, but that they shouldn't be easier to use to get access to the bad information. Can we just remove the "nor should" part, or does that lose something useful?

```suggestion
not make that information easier to access compared to existing comparable
web platform features.
```

index.html
```diff
@@ -1343,115 +1343,92 @@
...
+
+These APIs should also:
```
jyasskin, line 1405: This is just me being pedantic, so feel free to ignore if you don't want to add the words:

```suggestion
Specifications describing these APIs should also:
```

index.html
```diff
@@ -1343,115 +1343,92 @@
...
+
+* make it clear how to remove this access in the event that future web
+  platform changes make it possible to remove other access to the same
+  information.
+* make it clear how any [=user agent=] which blocks access to this kind of
+  information in general (perhaps by breaking some experiences on the web that
```
jyasskin, line 1411: It's actually for a user agent that blocks specifically the information that the new API would redundantly expose. Removing "in general" here might be enough, although we still have "this kind of information" which is broader than ideal.

```suggestion
  information (perhaps by breaking some experiences on the web that
```

Another approach might be around rephrasing the paragraph to talk about "A specification for a new API that redundantly exposes some undesirable information should: ... make it clear how any [=user agent=] whichblocks access to this information (perhaps...".

index.html
```diff
@@ -1343,115 +1343,92 @@
...
+  other browsers don't wish to break) can prevent the new API from exposing that
+  information without breaking additional sites or user experiences.
 
+<aside class="example">
+  Usually, these APIs will be designed to expose data that enables some
+  [=appropriate=] information discovery.
```
jyasskin, line 1417: Trying to make the web-without-3p-cookies reference flow better:

```suggestion
  Usually, these APIs will be designed to expose data that enables some
  [=appropriate=] information discovery, as recommended by [[[web-without-3p-cookies]]]:
  
  > It is better to approach [these use cases] with replacement technologies
  that are designed-for-purpose and built to respect user privacy.
```

jyasskin (approved):
I think this is good, once we've considered several comments. @yoavweiss, any concerns?

index.html
```diff
@@ -1343,115 +1343,92 @@
...
-to <a data-cite="design-principles#leave-the-web-better">leave the web better than you
-found it</a>.
-
-### Handling acceptable information {#acceptable-information}
-
 <div class="practice">
-  <span class="practicelab">New APIs can add ways of getting acceptable
-  information that are guarded at least as strongly as the existing ways.</span>
+  <span class="practicelab">Web APIs must not weaken [=access guards=] which
+  protect users from [=inappropriate=] exposure of their information.</span>
```
jyasskin, line 1348: Oh, right, I wanted to think about the principle itself after I read the whole new section. I originally wrote 3 principles because we'd figured out that there were 3 kinds of new APIs that got proposed, and each one implied a different set of constraints. You've covered the constraints in the section, and I'm not super concerned about whether we describe the constraints as principles or just in prose. The remaining question is how to summarize the point of the overall section, and I'm not sure this is actually the right principle for that...

Maybe it's more like:

```suggestion
  <span class="practicelab">New Web APIs must guard users' information at least
  as well as existing APIs that are expected to stay in the web platform.</span>
```

Or something about the "target state of the web platform", but "target state" might need too much explaining.

jyasskin: Or "at least as well as our concrete goals for the web platform"?

index.html
```diff
@@ -1343,115 +1343,92 @@
...
 API is acceptable because an existing acceptable API already exposes the same information,
 they must be careful to ensure that their new API is only available under a set of guards
-that's at least as strict. Without those guards, they need to make the argument from
+that are at least as strict. Without those guards, they need to make the argument from
 scratch, without relying on the existing API.
 
-### Handling information that's being removed {#unacceptable-information}
+If existing APIs provide access to some information, but there is a plan to
+change those APIs to prevent that access, new APIs must not be added that
+provide that same information, unless it is possible to put additional
```
yoavweiss, line 1377: Maybe a variant of "unless it is possible to include the new APIs as part of the same plan, and adding these new APIs won't make the plan harder"?

I'd also s/plan/plausible plan/

index.html
```diff
@@ -1343,115 +1343,92 @@
...
+  because the page can use <a data-cite="RFC9110#status.3xx">HTTP redirects</a> to
+  learn it, and redirection is a core feature of the web.
+
+  Some users are disappointed that a page with permission to run JavaScript can record
+  their pattern of interaction with that page. However, the page does this by using
+  the same events it would use to make the page interactive, so we can't block this
+  information access either.
+</aside>
+
+New APIs that unavoidably provide access to this kind of information should
```
yoavweiss, line 1401: I fundementally disagree with this section - if there's information with both bad and good uses, creating easier APIs to access it would hurt the good uses of that information while not necessarily impacting the bad uses (as they are typically more lucrative and deployed by folks that care less about the user experience).  E.g. based on this and the asides above, we wouldn't ship `ping` and IntersectionObserver. I don't think there's consensus that we regret shipping neither.

So a couple of caveats I'd like to see here:
* There needs to be consensus that the information is something we'd like to unship from the platform.
* The new APIs make it significantly cheaper (by an order of magnitude) to access that information.

index.html
```diff
@@ -1343,115 +1343,92 @@
...
-  we have a plan to change those APIs to prevent that access, new APIs must not be added
-  that provide that same information without extra [=access guards=] that make the access to the
-  information acceptable.</span>
-</div>
+For example, browsers are gradually removing the ability to join identities
+between different [=partitions=]. It is important that new APIs do not add
+features which re-enable [=cross-context recognition=].
 
-### Handling information we can't completely block {#sketchy-information}
+Some functionality of the web has historically been provided using features
```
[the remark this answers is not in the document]
chrisn: I think we should avoid the word "sketchy" (even as an anchor), it could be misunderstood by people where English is a second language, and in English it has multiple meanings - lacking detail / being misleading, neither of which really captures what's being described here.

index.html
```diff
@@ -1343,115 +1343,92 @@
...
-For example, the URLs of resources, the timing of link clicks, and the referrer chain within a
-single origin are not guarded by anything; the scroll position is guarded by the setting to turn off
-JavaScript; and access to the camera or geolocation are guarded by permission prompts.
+The many APIs available to websites expose lots of data that can be combined
+into information about people, web servers, and other things.
+
+User-controlled settings or permissions can <dfn data-lt="access guard">guard
+access</dfn> to data on the web. Whether or not exposing information is
+[=appropriate=] depends on the suitability of these [=access guards=] in
+[=context=].
```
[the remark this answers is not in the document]
rhiaro: I was unsure about swapping in `appropriate`, but I felt "acceptable" begs the question of: to whom? I think if we're going to use "acceptable" this way we should define it somewhere.

index.html
rhiaro: Fewer words always better!

(Held lightly; if you think it adds clarity, more words is better. I'll leave to you.)

index.html
rhiaro: I think the suggested edit is clear. I wouldn't want to complicate it by referring to "target state" or "concrete goals". This might want more noodling on by the group though, so I won't add the commit right now.

index.html
rhiaro: I think I understand the sense of "plausible" here, but it's potentially ambiguous/subjective. Perhaps something that leans on the standards process? A "concrete, consensus-based plan"? A "plan supported by multiple stakeholders on the web pltaform"?

index.html
```diff
@@ -1343,115 +1343,92 @@
...
-  that provide that same information without extra [=access guards=] that make the access to the
-  information acceptable.</span>
-</div>
+For example, browsers are gradually removing the ability to join identities
+between different [=partitions=]. It is important that new APIs do not add
+features which re-enable [=cross-context recognition=].
 
-### Handling information we can't completely block {#sketchy-information}
+Some functionality of the web has historically been provided using features
+that can be used to undermine people's privacy (e.g. third-party cookies). It
```
rhiaro, line 1385:
```suggestion
that can be used to undermine people's privacy. It
```

index.html
rhiaro, line 1384:
```suggestion
### Unavoidable information exposure

Some functionality of the web has historically been provided using features
```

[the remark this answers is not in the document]
rhiaro: I've suggested "Unavoidable information exposure" instead

rhiaro:
I've added some of the suggestions, left a few comments, and there are a few things to be resolved still (particularly @yoavweiss's comments). If the group discusses this today, I can do some more edits based on consensus later. Or, if it's easier for someone else to do the edits so it can be merged today, go ahead without me.

index.html
```diff
@@ -1343,115 +1343,97 @@
...
-user-controlled settings or permissions that <dfn data-lt="access guard">guard</dfn> access to it.
-For example, the URLs of resources, the timing of link clicks, and the referrer chain within a
-single origin are not guarded by anything; the scroll position is guarded by the setting to turn off
-JavaScript; and access to the camera or geolocation are guarded by permission prompts.
+The many APIs available to websites expose lots of data that can be combined
+into information about people, web servers, and other things.
+
+User-controlled settings or permissions can <dfn data-lt="access guard">guard
+access</dfn> to data on the web. Whether or not exposing information is
+[=appropriate=] depends on the suitability of these [=access guards=] in
```
jyasskin, line 1356:
```suggestion
access</dfn> to data on the web. Whether or not web APIs should expose information
depends on the suitability of these [=access guards=] in
```

index.html
npdoty: +1 to the clarity that these are additions to specifications, not APIs.

index.html
```diff
@@ -1343,115 +1343,97 @@
...
+[=context=].
```
jyasskin, line 1357:
```suggestion
access</dfn> to data on the web. When designing Web APIs, use access guards
to ensure they expose information in [=appropriate=] ways.
```

index.html
```diff
@@ -1343,115 +1343,97 @@
...
+
+<aside class="example">
+  For example, the URLs of resources, the timing of link clicks, and the referrer chain within a
+  single origin are not guarded by anything; the scroll position is guarded by the setting to turn off
```
hober, line 1361: Might be worth pointing out that even `` is also gated on JavaScript being enabled?

npdoty:
Condensed principle here might cover all three cases, but does the explanatory text cover all the different cases sufficiently?

index.html
jyasskin: `` is a good example of us following the overall principle in this section, but it doesn't really fit into this paragraph. Maybe a second paragraph:

> When the <{img/loading}>=<{img/loading/lazy}> attribute was added, the designers realized that it exposed the scroll position, so it's also guarded by the setting to turn off JavaScript.

I'm not sure those [linking shorthands](https://speced.github.io/bikeshed/#autolink-wrappers) are supported by Respec...

index.html
```diff
@@ -1343,115 +1343,96 @@
...
+into information about people, web servers, and other things.
+
+User-controlled settings or permissions can <dfn data-lt="access guard">guard
+access</dfn> to data on the web. When designing a Web API, use [=access guards=]
+to ensure the API exposes information in [=appropriate=] ways.
+
+<aside class="example">
+  For example, the URLs of resources, the timing of link clicks, and the referrer chain within a
+  single origin are not guarded by anything; the scroll position is guarded by the setting to turn off
+  JavaScript; and access to the camera or geolocation are guarded by permission prompts.
```
jyasskin, line 1361:
```suggestion
  For example, the URLs of resources, the timing of link clicks, and the referrer chain within a
  single origin are not guarded by anything; the scroll position is guarded by the setting to turn off
  JavaScript; and access to the camera or geolocation are guarded by permission prompts.
  
  When the `<img loading=lazy>` attribute was added, the designers realized
  that it exposed the scroll position, so it's also guarded by the setting to turn
  off JavaScript.
```

index.html
jyasskin: FWIW, I think this is a comment about the original text, rather than this change, and we should sit down once this PR is merged and try to fix up the whole section to match your understanding better. I don't think it should block merging this editorial PR.

PR #221, opened 2023-02-23
remove list of privacy labor suggestions around ancillary data

pes10k:
Remove list of suggestions around privacy labor and ancillary data. 

I also think it'd be fine to solve #220 by being more specific about how browsers could prevent privacy labor around ancillary data collection w/o collecting ancillary data from users who dont want to or anticipate that data collection

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/pull/221.html#ancillary-uses" title="Last updated on May 11, 2023, 12:45 AM UTC (e2aa4bd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/221/d52ee70...e2aa4bd.html" title="Last updated on May 11, 2023, 12:45 AM UTC (e2aa4bd)">Diff</a>

PR #359, opened 2023-09-26
Rewrite the ancillary data section.

### LINKED ISSUE #220: Remove or revise examples of signals browsers should use to prevent "privacy labor" in ancillary data
"Section 2.2.1 Ancillary uses" gives several examples of things browsers may use to decide when / what ancillary data to send.

Some of the mentioned items here clash with the previous paragraphs, which specifically call out that some users wont want any of this. Some of the examples here are specifically not about the current user (i.e., "user research") and so seem very likely to lead the browser to send data the current user (in some cases) are unaware of and / or don't want to participate in.

Using "heuristics of sensitivity of data" also seems likely to lead to user harm, since, as section "2.4 Sensitive Information" notes, "A particular piece of information may have different sensitivity for different people."  We should not be encouraging browsers to collect possibly sensitive data for purposes not intended to immediately benefit users unless we have signals from the current user that its not sensitive to the current user. Unless we can be more specific about what heuristics are meant here, this seems likely to harm at least some users

jyasskin:
Hopefully this satisfies both the WebPerf and Privacy goals...

Sorry to Amy for probably undoing their readability fixes.  🫣

Fixes #220.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/359.html#ancillary-uses" title="Last updated on Sep 28, 2023, 4:36 PM UTC (df6b6c1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/359/adf2101...jyasskin:df6b6c1.html" title="Last updated on Sep 28, 2023, 4:36 PM UTC (df6b6c1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1124,14 +1124,14 @@ the data exposed, is to autogenerate a new email address for each context.
 
 <div class="practice" data-audiences="api-designers">
 <span class="practicelab"
-id="practice-ancillary-data-shouldnt-reveal-personal-data">APIs that expose
+id="principle-ancillary-data-shouldnt-reveal-personal-data">APIs that expose
 [=ancillary data=] should not reveal any [=personal data=] that isn't already
 available through other APIs, without the identifiable person's
 permission.</span>
 </div>
 
 <div class="practice" data-audiences="user-agents">
-<span class="practicelab" id="practice-disabling-ancillary-data">User agents
+<span class="practicelab" id="principle-disabling-ancillary-data">User agents
 should provide a way to disable APIs that expose [=ancillary data=].</span>
 </div>
 
--- a/index.html
+++ b/index.html
@@ -1122,37 +1122,84 @@ the data exposed, is to autogenerate a new email address for each context.
 
 ### Ancillary uses
 
-In order to uphold the principle of [[[#data-minimization]]], [=sites=] and
-[=user agents=] should seek to understand and respect people's goals and preferences about
-use of data about them.
+<div class="practice" data-audiences="api-designers">
+<span class="practicelab"
+id="practice-ancillary-data-shouldnt-reveal-personal-data">APIs that expose
+[=ancillary data=] should not reveal any [=personal data=] that isn't already
+available through other APIs, without the identifiable person's
+permission.</span>
+</div>
+
+<div class="practice" data-audiences="user-agents">
+<span class="practicelab" id="practice-disabling-ancillary-data">User agents
+should provide a way to disable APIs that expose [=ancillary data=].</span>
+</div>
 
 [=Sites=] sometimes use data in ways that aren't needed for the user's immediate
-goals. These uses are known as <dfn data-lt="ancillary use">ancillary uses</dfn>,
-and data that is primarily useful for [=ancillary uses=] is <dfn>ancillary data</dfn>.
+goals. For example, they might bill advertisers, measure site performance, or
+alert developers to bugs. These uses are known as <dfn data-lt="ancillary
+use">ancillary uses</dfn>. Data that is primarily useful for [=ancillary uses=]
+is <dfn>ancillary data</dfn>.
 
-<aside class="example">
-  Some examples of [=ancillary data=] include data used for browser telemetry, site telemetry,
-  performance measurements, and software updates.
+Some data that is exposed for non-[=ancillary uses=] can also be used for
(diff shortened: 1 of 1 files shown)


npdoty:
#220 is an issue about revising the examples in this section, and #221 is a PR that would address that issue.

Is this re-write addressing some other issue? Or a series of issues, given how completely it changes the existing text?

jyasskin:
Unfortunately, #221 created a conflict with the WebPerf WG. The old text appeared to say that if an API was primarily good for ancillary uses, then it was producing ancillary data, which meant most of the WebPerf APIs produced ancillary data. The [example](https://web.archive.org/web/20230927225121/https://w3ctag.github.io/privacy-principles/#example-7) didn't help, since it didn't mention any concrete data. In the cases where WebPerf were just summarizing data produced by other web APIs, they didn't think it was reasonable to "aggressively minimize" the data without being able to use research or heuristics to decide some of it was ok.

@yoavweiss pointed out that it works better to focus on the source of the data, and less on how it's processed or summarized. He pointed out particular reporting APIs that expose new ancillary data, and others that merely summarize non-ancillary data that's available from other sources. That let me improve the [example](https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/359.html#example-ancillary-data) to name concrete telemetry APIs that do or don't return ancillary data. It also let me lay out a decision tree for API designers and reviewers to follow in order to figure out how to treat their new API. Listing the concrete options for API designers did take more space than generally describing what users might want, but I think it's more useful too.

npdoty:
Could we open an issue that describes the concerns that WebPerf would have, with either the existing text or with the text in #221?

My understanding is that performance analysis was widely agreed to be an ancillary use of data. And that https://www.w3.org/TR/privacy-principles/#information was specifically written, directly after conversations with WebPerf folks, to address the distinctions between data available from other/existing APIs.

jyasskin:
I've asked @yoavweiss to organize an issue or a comment on the existing issues from WebPerf's perspective, and he thinks he'll get to that tomorrow.

Performance analysis is an ancillary use, but that doesn't make all of its inputs into ancillary data.

The [Ancillary uses](https://www.w3.org/TR/privacy-principles/#ancillary-uses) section seems to be trying to add restrictions on top of https://www.w3.org/TR/privacy-principles/#information. If it weren't adding restrictions, we could delete the whole section. But I think a few additional restrictions are useful, and that the [principles I've added in this PR](https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/359.html#principle-ancillary-data-shouldnt-reveal-personal-data) will lead to concrete privacy improvements in some of WebPerf's APIs.

Your original text had to be somewhat vague about this because we hadn't yet figured out how the WebPerf folks were thinking about their designs. Now that we have the distinction that some of their APIs are streamlining existing uses of APIs, while others are providing new capabilities, we can make the text more concrete. We should be willing to make more than local tweaks in that direction.

yoavweiss:
Thanks @jyasskin!! (and apologies for the delay)

From the WebPerfWG's perspective, the previous principles were vague in ways that made their application open to interpretation, which I felt could lead to future disagreements.

IMO, the issue stems from the definition of ancillary data based on its "primary" use, regardless of the potential reasons to why the data is non-ancillarily exposed. Jeffrey's modifications on that front are very much welcome, as they create a clear deliniating line between data that is already exposed for functional reasons, and data that is *only* exposed for ancillary uses.

That kind of principle can enable us at the WebPerfWG to clearly mark the APIs we're working on and their different attributes as "ancillary" vs. "non-ancilllary", and think of ways in which we can work on migrating the ancillary data to safer options (e.g. aggregated exposure) without harming the use cases it tackles.

jyasskin:
Major discussion in https://github.com/w3ctag/privacy-principles/blob/main/meetings/2023-10-04-minutes.md, which I'll try to apply to a new PR.

pes10k:
@jyasskin adding my three comments here, though happy to move them to the follow up PR too if they're applicable

- one principal calls out "without the identifiable person's permission".  I wasn't sure what work "identifiable" is doing here, and if its meant to distinguish different categories of "personal data"
- CSP reports are listed as a type of report that summarizes existing data, but this isn't 100% correct.  CSP reports include some information not currently easily accessible by the page (e.g., the violating script's text)
- I wasn't sure which principal the "this principal" text on line 1178 is referring to

PR #361, opened 2023-10-06
Rewrite ancillary uses to focus on 2 kinds of ancillary APIs rather than ancillary data.

### LINKED ISSUE #220: Remove or revise examples of signals browsers should use to prevent "privacy labor" in ancillary data

jyasskin:
This is a replacement for #359 after the discussion in https://github.com/w3ctag/privacy-principles/blob/main/meetings/2023-10-04-minutes.md. It will close #220.

@pes10k, I think I fixed your [comments](https://github.com/w3ctag/privacy-principles/pull/359#issuecomment-1747625487):
* "identifiable person" was an echo of a phrase in the [definition of "personal data"](https://w3ctag.github.io/privacy-principles/#dfn-data), but I found a way to avoid it.
* I replaced the CSP report example with event timings. There's also a nit there that the [duration](https://www.w3.org/TR/event-timing/#priv-sec) field is a novel piece of data, so I could narrow this to just the [`processingEnd`](https://www.w3.org/TR/event-timing/#dom-performanceeventtiming-processingend) field.
* I reworded the "it complies with this principle" bit and also moved the second principle in the section down to just before the paragraph that elaborates on it. This violates the guideline we've been following to put principles at the tops of their sections, so maybe folks will want me to move it back.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/361.html#ancillary-uses" title="Last updated on Nov 15, 2023, 6:45 PM UTC (b2e5a25)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/361/7dd3a4f...jyasskin:b2e5a25.html" title="Last updated on Nov 15, 2023, 6:45 PM UTC (b2e5a25)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1260,12 +1260,9 @@ across contexts</a>. In order to uphold the principle of [[[#data-minimization]]
 [=user agents=] should seek to understand and respect people's goals and preferences about
 use of this data.
 
-[=User agents=] usually can't do much about data collection from the
-[=non-ancillary APIs=].
-
-There is disagreement about how [=user agents=] should handle [=ancillary APIs
-computed from existing information=]. Advocates of these APIs argue that they're
-hard to use to
+The task force does not have consensus about how [=user agents=] should handle
+[=ancillary APIs computed from existing information=].
+Advocates of these APIs argue that they're hard to use to
 extract [=personal data=], they're more efficient than collecting the same
 information though [=non-ancillary APIs=], sites are less likely to adopt these
 APIs if a significant number of people turn them off, and that the act of
@@ -1275,8 +1272,7 @@ collect it, and because there's still some risk, users should be able
 to turn off this group of APIs that probably won't directly break a site's
 functionality.
 
-We do have consensus that [[[#information]]] governs [=ancillary APIs computed
-from existing information=] and that:
+Because different users are likely to have different preferences:
 
 <div class="practice" data-audiences="api-designers">
 <span class="practicelab" id="principle-identify-ancillary-apis">Specifications
@@ -1291,7 +1287,8 @@ agents=] can provide appropriate choices for their users.</span>
 <span class="practicelab"
 id="principle-ancillary-apis-with-new-information-shouldnt-reveal-personal-data">
 [=Ancillary APIs that provide new information=] should not reveal any [=personal
-data=] that isn't already available through other APIs, without permission.
+data=] that isn't already available through other APIs, without an indication
+that doing so aligns with the user's wishes and interests.
 </span>
 </div>
 
@@ -1331,8 +1328,9 @@ few choices in this situation:
   ask the user whether they generally support sharing this data, rather than
   asking for each use of the APIs.
 
-API designers should maintain APIs that had to make one of these choices and
-should keep trying to evolve them toward aggregating the data instead.
+If an API had to make one of these choices, and then something else about the
+API needs to change, designers should consider replacing the whole API with one
+that avoids exposing [=personal data=].
 
 Some other [=ancillary uses=] do require that a person be connected to their
 data. For example, a person might want to file a bug report that a website
--- a/index.html
+++ b/index.html
@@ -1237,8 +1237,8 @@ use">ancillary uses</dfn>.
     APIs that filter, summarize, or time-shift information available from
     [=non-ancillary APIs=], like the [[[event-timing]]] and <a
     data-cite="intersection-observer#introduction">IntersectionObserver</a>. See
-    [[#information]] for restrictions on how existing non-ancillary APIs can be
-    used to justify new ancillary APIs.
+    [[[#information]]] for restrictions on how existing non-ancillary APIs can
+    be used to justify new ancillary APIs.
   </dd>
 
   <dt><dfn>Ancillary APIs that provide new information</dfn></dt>
--- a/index.html
+++ b/index.html
@@ -1234,8 +1234,8 @@ use">ancillary uses</dfn>.
 
   <dt><dfn>Summarizing ancillary APIs</dfn></dt>
   <dd>
-    APIs that filter or summarize information available from [=non-ancillary
-    APIs=], like the [[[event-timing]]] and <a
+    APIs that filter, summarize, or time-shift information available from
+    [=non-ancillary APIs=], like the [[[event-timing]]] and <a
     data-cite="intersection-observer#introduction">IntersectionObserver</a>.
   </dd>
 
--- a/index.html
+++ b/index.html
@@ -1216,37 +1216,91 @@ the data exposed, is to autogenerate a new email address for each context.
 
 ### Ancillary uses
 
-In order to uphold the principle of [[[#data-minimization]]], [=sites=] and
-[=user agents=] should seek to understand and respect people's goals and preferences about
-use of data about them.
-
 [=Sites=] sometimes use data in ways that aren't needed for the user's immediate
-goals. These uses are known as <dfn data-lt="ancillary use">ancillary uses</dfn>,
-and data that is primarily useful for [=ancillary uses=] is <dfn>ancillary data</dfn>.
+goals. For example, they might bill advertisers, measure site performance, or
+tell developers about bugs. These uses are known as <dfn data-lt="ancillary
+use">ancillary uses</dfn>.
 
-<aside class="example">
-  Some examples of [=ancillary data=] include data used for browser telemetry, site telemetry,
-  performance measurements, and software updates.
-</aside>
+[=Sites=] can get the data they want for [=ancillary uses=] from a variety of places:
 
-Different [=users=] will want to share different kinds and amounts of
-[=ancillary data=] with [=sites=]. Some [=people=] will not want to share any
-[=ancillary data=] at all.
+<dl>
+  <dt><dfn>Non-ancillary APIs</dfn></dt>
+  <dd>
+    Web APIs that were designed to support users' immediate goals, like <a
+    data-cite="dom#interface-event">DOM events</a> and <a
+    data-cite="cssom-view-1#extension-to-the-element-interface">element position
+    observers</a>.
+  </dd>
 
-Users may be willing to share [=ancillary data=] if it is aggregated with
-the data of other users, or [=de-identified=]. This can be useful
-when [=ancillary data=] contributes to a collective benefit in a way
-that reduces privacy threats to individuals (see <a href="#principle-collective-privacy">collective
-privacy</a>).
+  <dt><dfn>Summarizing ancillary APIs</dfn></dt>
+  <dd>
+    APIs that filter or summarize information available from the first group of
+    APIs, like the [[[event-timing]]] and <a
+    data-cite="intersection-observer#introduction">IntersectionObserver</a>.
+  </dd>
 
-<aside class="example">
-  Privacy-preserving measurement techniques may be used for aggregate calculations while minimizing
-  the number of actors that have access to personal data about many individual people. Encryption and
-  privacy-preserving proxies may minimize the number of actors that have access to personal data or
-  hide the contents of personal data. But even
-  with those protections, some people may prefer not to participate in some kinds of measurement.
+  <dt><dfn>Novel ancillary APIs</dfn></dt>
+  <dd>
+    APIs that provide new information that's primarily useful to support the
+    ancillary uses, like <a data-cite="element-timing#sec-intro">element paint
+    timing</a>, <a data-cite="performance-measure-memory#intro">memory usage
+    measurements</a>, and <a
+    data-cite="deprecation-reporting#deprecation-report">deprecation
+    reports</a>.
+  </dd>
+</dl>
 
-  There is ongoing work on these kinds of technologies in the <abbr title="Internet Engineering Task
+All of these sources of data can reveal [=personal data=] about a person's
+configuration, device, environment, or behavior that could be <a
+href="#hl-sensitive-information">sensitive</a> or be used as part of <a>browser
+fingerprinting</a> to <a data-lt="cross-context recognition">recognize people
+across contexts</a>. In order to uphold the principle of [[[#data-minimization]]], [=sites=] and
+[=user agents=] should seek to understand and respect people's goals and preferences about
+use of this data.
+
+[=User agents=] usually can't do much about data collection from the
+[=non-ancillary APIs=].
+
+There is disagreement about how [=user agents=] should handle [=summarizing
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1122,37 +1122,91 @@
...
+  </dd>
 
-Different [=users=] will want to share different kinds and amounts of
-[=ancillary data=] with [=sites=]. Some [=people=] will not want to share any
-[=ancillary data=] at all.
+  <dt><dfn>Summarizing ancillary APIs</dfn></dt>
+  <dd>
+    APIs that filter or summarize information available from [=non-ancillary
+    APIs=], like the [[[event-timing]]] and <a
+    data-cite="intersection-observer#introduction">IntersectionObserver</a>.
```
yoavweiss, line 1145: I wouldn't consider intersection observer ancillary. It's somewhat polyfillable, but can be used to e.g. lazily load parts of the UI.

index.html
```diff
@@ -1122,37 +1122,91 @@
...
+    observers</a>.
+  </dd>
 
-Different [=users=] will want to share different kinds and amounts of
-[=ancillary data=] with [=sites=]. Some [=people=] will not want to share any
-[=ancillary data=] at all.
+  <dt><dfn>Summarizing ancillary APIs</dfn></dt>
+  <dd>
+    APIs that filter or summarize information available from [=non-ancillary
+    APIs=], like the [[[event-timing]]] and <a
```
yoavweiss, line 1144: s/the//

index.html
```diff
@@ -1122,37 +1122,91 @@
...
+    data-cite="cssom-view-1#extension-to-the-element-interface">element position
+    observers</a>.
+  </dd>
 
-Different [=users=] will want to share different kinds and amounts of
-[=ancillary data=] with [=sites=]. Some [=people=] will not want to share any
-[=ancillary data=] at all.
+  <dt><dfn>Summarizing ancillary APIs</dfn></dt>
+  <dd>
+    APIs that filter or summarize information available from [=non-ancillary
```
yoavweiss, line 1143: Maybe "filter, summarize or time-shift"?
E.g. ResourceTiming reports (almost) every resource in ways that can be polyfilled by patching `fetch()`, XHR and walking the DOM/mutation observers. It time-shifts the reports (a little bit) and ensures their handling is async, and not on the critical path of the resource.

index.html
```diff
@@ -1122,37 +1122,91 @@
...
+    APIs=], like the [[[event-timing]]] and <a
+    data-cite="intersection-observer#introduction">IntersectionObserver</a>.
+  </dd>
 
-Users may be willing to share [=ancillary data=] if it is aggregated with
-the data of other users, or [=de-identified=]. This can be useful
-when [=ancillary data=] contributes to a collective benefit in a way
-that reduces privacy threats to individuals (see <a href="#principle-collective-privacy">collective
-privacy</a>).
+  <dt><dfn>Novel ancillary APIs</dfn></dt>
```
yoavweiss, line 1148: Maybe "new sources of ancillary data"?

index.html
```diff
@@ -1122,37 +1122,91 @@
...
+[=non-ancillary APIs=].
+
+There is disagreement about how [=user agents=] should handle [=summarizing
+ancillary APIs=]. Advocates of these APIs argue that they're hard to use to
+extract [=personal data=], they're more efficient than collecting the same
+information though [=non-ancillary APIs=], sites are less likely to adopt these
+APIs if a significant number of people turn them off, and that the act of
+turning them off can ironically contribute to [=browser fingerprinting=].
+Opponents argue that if data's easier or cheaper to collect, more sites will
+collect it, and because there's still some risk and cost, users should be able
```
yoavweiss, line 1177: Is "cost" here related to privacy?

jyasskin: Maybe: it's about the computational and network costs, so not about exposing user information, but I think the connection to privacy is that it's about using a user's resources without their knowledge for things they might not like. But the closest item in https://wiki.openrightsgroup.org/wiki/A_Taxonomy_of_Privacy that seems to match is Intrusion, but that's limited to when the intrusion is [“highly offensive to a reasonable person”](https://scholarship.law.gwu.edu/cgi/viewcontent.cgi?article=2074&context=faculty_publications#page=74), and that's debatable for these APIs.

@pes10k, I'm trying to represent your concern with this bit, so maybe you can justify why this is a privacy concern better than I can?

index.html
jyasskin: This renders as "like the Event Timing API", so the "the" is useful.

index.html

index.html
jyasskin: I could do that or 

```suggestion
  <dt><dfn>Ancillary APIs that provide new information</dfn></dt>
```

along with `<dfn>Ancillary APIs that filter or summarize other sources of information</dfn>` above to stay parallel. I'll give other folks a chance to weigh in before making this change.

index.html
```diff
@@ -1161,34 +1215,43 @@
...
+  that a person has a fast graphics card, that they click slowly, or that they
+  use a certain proxy, but the fact that they click slowly is already revealed
+  by <a data-cite="dom#interface-event">DOM event</a> timing.
+* [=User agents=] can ask their users' permission to enable this class of API.
+  To reduce [=privacy labor=], a [=user agent=] could use a first-run dialog to
+  ask the user whether they generally support sharing this data, rather than
+  asking for each use of the APIs.
+
+API designers should maintain APIs that had to make one of these choices and
+should keep trying to evolve them toward aggregating the data instead.
```
jyasskin, line 1236: @pes10k @sandandsnow You'd been concerned about saying that "aggregation" is sufficient for depersonalizing data. Is there a term I should use instead? Above I used "private aggregation", but I'm happy to use something else or to define something.

index.html
```diff
@@ -1122,37 +1122,91 @@
...
+information though [=non-ancillary APIs=], sites are less likely to adopt these
+APIs if a significant number of people turn them off, and that the act of
+turning them off can ironically contribute to [=browser fingerprinting=].
+Opponents argue that if data's easier or cheaper to collect, more sites will
+collect it, and because there's still some risk and cost, users should be able
+to turn off this group of APIs that probably won't directly break a site's
+functionality.
+
+We do have consensus that [[[#information]]] governs [=summarizing ancillary
+APIs=] and that:
```
npdoty, line 1182: Do we need to note consensus that the other principles in the document still apply? That would seem to be the case for all principles.

index.html
```diff
@@ -1161,34 +1215,43 @@
...
 
-[=User agents=] should aggressively <a href="#data-minimization">minimize</a> [=ancillary
-data=] and should avoid burdening the user with additional [=privacy labor=]
-when deciding what [=ancillary data=] to expose. To that end, user agents may
-employ user research, solicitation of general preferences, and heuristics about
-sensitivity of data or trust in a particular [=context=].
-
-To help [=sites=] understand user preferences, user agents can provide
-browser-configurable signals to directly communicate common user preferences
-(such as a [=global opt-out=]).
```
npdoty, line 1172: why is this being deleted?

index.html
```diff
@@ -1161,34 +1215,43 @@
   Group">PATCG</abbr></a> groups.
 </aside>
 
-[=User agents=] should aggressively <a href="#data-minimization">minimize</a> [=ancillary
-data=] and should avoid burdening the user with additional [=privacy labor=]
```
npdoty, line 1165: why remove the requirement to minimize this data?

index.html
```diff
@@ -1161,34 +1215,43 @@
...
+  data=], as described by [[[#information]]]. For example, the API might reveal
+  that a person has a fast graphics card, that they click slowly, or that they
+  use a certain proxy, but the fact that they click slowly is already revealed
+  by <a data-cite="dom#interface-event">DOM event</a> timing.
+* [=User agents=] can ask their users' permission to enable this class of API.
+  To reduce [=privacy labor=], a [=user agent=] could use a first-run dialog to
+  ask the user whether they generally support sharing this data, rather than
+  asking for each use of the APIs.
+
+API designers should maintain APIs that had to make one of these choices and
```
npdoty, line 1235: what is the maintenance requirement? I'm not clear what we're asking designers to do.

index.html
npdoty, line 1236: Is this a useful exhortation? What would convince API designers to migrate to an aggregation alternative if they aren't required to minimize data and users are actively discouraged from disabling this functionality?

npdoty (dismissed):
I'm not clear what issue this addresses or how this re-write improves on the existing text.

This seems to repeat and more deeply encode the new-vs-old distinction, which seems to set up further debates over how minimization shouldn't apply if something is old enough or has some alternative use.

This seems to delete much of the text about solicitation and respecting of general preferences around analytics and other ancillary uses. Further, it seems to actively discourage browsers from providing general choices.

index.html
pes10k: (fwiw, @npdoty expressed the concern i was trying to get across)

index.html
jyasskin: @npdoty suggested using the ["unavoidable information exposure"](https://w3ctag.github.io/privacy-principles/#unavoidable-information-exposure) phrase, and I'll try to work that in.

index.html
jyasskin: It's not related to ancillary data, and https://w3ctag.github.io/privacy-principles/#dfn-global-opt-out still says that UAs can provide this sort of signal.

index.html
jyasskin: API designers are required to minimize data by https://w3ctag.github.io/privacy-principles/#data-minimization. I'd be ok with deleting this paragraph if folks don't think it's useful, but I figured it'd be useful to encourage designers to keep trying to find alternatives that don't reveal personal data. We could also re-word it if there's a clearer wording for that.

index.html
```diff
@@ -1161,34 +1215,43 @@
...
+Some other [=ancillary uses=] do require that a person be connected to their
+data. For example, a person might want to file a bug report that a website
+breaks on their particular computer, and be able to get follow-up communication
+from the developers while they fix the bug. This is an appropriate time to ask
+the person's permission.
+
+<div class="practice" data-audiences="user-agents">
+<span class="practicelab" id="principle-disabling-novel-ancillary-apis">User
+agents should provide a way to disable [=novel ancillary APIs=].</span>
+</div>
```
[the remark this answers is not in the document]
jyasskin: The other ancillary APIs aren't providing "very similar data". If they were, this set of APIs wouldn't "provide new information."

UAs are also free to have their setting turn off more APIs than the ones called out here; this just sets a minimum bar.

index.html
jyasskin: I'd be fine with dropping that.

index.html
```diff
@@ -1122,37 +1122,91 @@
...
+
+[=User agents=] usually can't do much about data collection from the
+[=non-ancillary APIs=].
+
+There is disagreement about how [=user agents=] should handle [=summarizing
+ancillary APIs=]. Advocates of these APIs argue that they're hard to use to
+extract [=personal data=], they're more efficient than collecting the same
+information though [=non-ancillary APIs=], sites are less likely to adopt these
+APIs if a significant number of people turn them off, and that the act of
+turning them off can ironically contribute to [=browser fingerprinting=].
```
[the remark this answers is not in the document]
jyasskin: I've removed "ironically".

index.html
```diff
@@ -1122,37 +1122,91 @@
...
+fingerprinting</a> to <a data-lt="cross-context recognition">recognize people
+across contexts</a>. In order to uphold the principle of [[[#data-minimization]]], [=sites=] and
+[=user agents=] should seek to understand and respect people's goals and preferences about
+use of this data.
+
+[=User agents=] usually can't do much about data collection from the
+[=non-ancillary APIs=].
+
+There is disagreement about how [=user agents=] should handle [=summarizing
+ancillary APIs=]. Advocates of these APIs argue that they're hard to use to
```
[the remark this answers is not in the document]
jyasskin: I'd be fine with removing this paragraph if that's what the group wants to do.

index.html
```diff
@@ -1122,37 +1122,91 @@
...
+All of these sources of data can reveal [=personal data=] about a person's
+configuration, device, environment, or behavior that could be <a
+href="#hl-sensitive-information">sensitive</a> or be used as part of <a>browser
+fingerprinting</a> to <a data-lt="cross-context recognition">recognize people
+across contexts</a>. In order to uphold the principle of [[[#data-minimization]]], [=sites=] and
+[=user agents=] should seek to understand and respect people's goals and preferences about
+use of this data.
+
+[=User agents=] usually can't do much about data collection from the
+[=non-ancillary APIs=].
```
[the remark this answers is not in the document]
jyasskin: It's mostly here because I wanted to say something about all three categories that I introduced above. I'm fine with removing it.

index.html
```diff
@@ -1216,37 +1216,95 @@
...
+
+There is disagreement about how [=user agents=] should handle [=ancillary APIs
```
jyasskin, line 1266: From @torgo:
```suggestion
The task force does not have consensus about how [=user agents=] should handle [=ancillary APIs
```

index.html
jyasskin: Reworded to

> Because different users are likely to have different preferences:

That's still a little awkward, so I'd welcome other wording suggestions.

pes10k:
From slack

```I think this change looks good to me, though I also was interested in @torgo re-bringing-up the idea of moving back to the for-the-user/for-the-site way of defining ancillary data.  I think I was only the main person supporting that direction , and so stopped pushing it.

But if @torgo (or others) supports that direction again, as an alternative to the “yes/no exists and not planned for removal in existing APIs” way of splitting things up, then I’d like to revist```

pes10k (approved):

torgo (approved):
we agreed to merge on today's taskforce call

PR #363, opened 2023-10-31
Don't say collective decision-making is strictly superior to engagement with individuals.

### LINKED ISSUE #351: "should best" and collective privacy
> [Principle](https://w3ctag.github.io/privacy-principles/#principle-collective-privacy): Groups and various forms of institutions should best protect and support autonomy by making decisions collectively rather than individually to either prevent or enable data sharing, and to set defaults for data processing rules.

I'm a little unclear on the grammatical intent of the phrasing of this statement, particularly around the "should best" part.

However, I think that this is making too strong a statement.  There are cases where collective action on privacy, or collective setting of rules, has a great deal of power and utility.  @jyasskin gave a great example in the TPAC session here that highlights how standards development is a good way to use collective decision-making processes to achieve positive privacy outcomes, particularly for cases where the issues are complex.

However, the way that this is phrased could be read to imply that collective setting of rules is strictly superior to engagement with individuals.  I'm guessing that that is unintentional, so I suggest that some editorial work is needed to clarify the intent if that is the case.  Otherwise, I think we might want to debate the point.

jyasskin:
Attempts to fix #351 using @npdoty's [wording](https://github.com/w3ctag/privacy-principles/issues/351#issuecomment-1768760294).

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/363.html#collective-privacy" title="Last updated on Oct 31, 2023, 9:39 PM UTC (08378cd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/363/5853646...jyasskin:08378cd.html" title="Last updated on Oct 31, 2023, 9:39 PM UTC (08378cd)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1494,8 +1494,8 @@ Note that [=controlled de-identified data=], on its own, is not sufficient to ma
 
 <p>
   <span class="practicelab" id="principle-collective-privacy">
-    Groups and various forms of institutions should best protect and support autonomy by making
-    decisions collectively rather than individually to either prevent or enable data
+    Groups and institutions should support autonomy by making
+    decisions collectively to either prevent or enable data
     sharing, and to set defaults for data processing rules.
   </span>
 </p>


index.html
```diff
@@ -1494,8 +1494,8 @@
 
 <p>
   <span class="practicelab" id="principle-collective-privacy">
-    Groups and various forms of institutions should best protect and support autonomy by making
-    decisions collectively rather than individually to either prevent or enable data
+    Groups and institutions should support autonomy by making
+    decisions collectively to either prevent or enable data
     sharing, and to set defaults for data processing rules.
```
martinthomson, line 1499: I'm not really sure that this says what it should now.  The right words are hard to get here, but there are few different things that you might say:

1. Institutions need to consider individual autonomy as a goal when making decisions about how to structure the systems they produce, whether it be regulation or voluntary standards.
2. Institutions need to consider the potential for individuals to have diverse goals as it relates to autonomy, rather than thinking of individuals as a uniform collective, and account for those diverse goals when making decisions [etc...].
3. In some settings, decisions made collectively for a population can provide the best ability to ensure that individual autonomy is respected, particularly where actions by a single individual might otherwise have stronger adverse consequences for privacy, such as retaliation or becoming more identifiable as a result of that action.
4. Collective decision-making can provide better outcomes where decisions need to carefully balance individual autonomy against collective harms or benefits.

I'm sure that there are more and more nuanced things to be said here, but I don't know which of these applies.  Perhaps all of them and more.

npdoty: These are useful points to make in the longer explanatory text of this section. There are some existing examples for when collective decision-making can facilitate individual autonomy (like when the decisions are too frequent or complex for an individual), but you should suggest some additional conditions.

Maybe we could merge the fix to the principle, and then Martin you could propose either longer text for this section, or some separate piece of writing that we could cite.

darobin: I don't disagree that there's considerable potential for subtlety and nuance in how institutions relate to individual autonomy. However, I am concerned that in trying to get into too much detail here exceeds the bounds of our legitimate expertise. Doing it right would also extend the text by a lot, and in a way that would not alleviate concerns made by some (including @martinthomson :) that this document is already academic AF.

darobin (approved):

torgo (approved):
lgtm - as is and we discussed that we can build on this with additional text on collective decision making as part of the outcome of #351.

PR #365, opened 2023-11-01
Copy-edit the Collective Privacy section.

### LINKED ISSUE #308: Copy edit section 2.7, Collective Privacy

jyasskin:
I significantly simplified and shortened the wording, so please check that this still says what we want it to say.

This includes the change to the principle in #363, and updates to that PR might imply that we need some other changes to this one.

This should fix #308.


index.html
```diff
@@ -1504,57 +1504,69 @@
...
+which lets a user agent either implement it quickly and asynchronously
+or ignore it for users who want to keep that information private.
+
+Unfortunately, while this arrangement did improve performance,
+it turned out not to help privacy.
+What prevents a site from using [^a/ping^] for [=people=] who have it activated
+and bounce tracking for others? What prevents a browser from opting everyone out because it wishes
 to offer better protection by default? Given the contested nature of the [^a/ping^] attribute and
 the absence of a forcing function to support collective enforcement, the scheme failed to deliver
 improved privacy.
```
jyasskin, line 1566: I wasn't sure exactly what we meant here by "a forcing function to support collective enforcement", so I haven't changed it to simpler wording. What should get enforced? That websites exclusively use `ping` instead of bounce tracking?

I'm also not sure that we're right that `<a ping>` didn't help privacy. My understanding is that [there are tracking-blocking extensions that do block the network requests from the `ping` attribute](https://github.com/gorhill/uBlock/wiki/Dashboard:-Settings#disable-hyperlink-auditing), and that websites have some trouble falling back to bounce tracking in that case. Is there something we can cite that'll better demonstrate that there's a problem?

index.html
```diff
@@ -1504,57 +1504,69 @@
...
+
+  <li>
+    When the [=data processing=] is justified or unjustified at a societal level,
+    even accounting for individual disagreements ([[?Relational-Governance]]).
+  </li>
+</ul>
+
+Different forms of collective decision-making are legitimate depending on what data is being processed.
+These forms might be governmental bodies at various administrative levels, standards
 organisations, worker bargaining units, or civil society fora.
```
darobin, line 1538: Not vital, but wondering if we should reference: https://www.mnot.net/blog/2023/11/01/regulators

index.html
```diff
@@ -1504,57 +1504,69 @@
...
-basis — even if the [=processing=] is relatively simple — is unrealistic if we also want them to have
-reasonable levels of [=autonomy=] in making these decisions.
-
-The purpose of this principle is to require that [=data governance=] provide ways to distinguish
-[=appropriate=] [=data=] [=processing=] without relying on individual decisions whenever the latter
-are impossible, which is often ([[?Relational-Governance]], [[?Relational-Turn]]).
-
-Which forms of collective governance are recognised as legitimate will depend on domains. These may
-take many forms, such as governmental bodies at various administrative levels, standards
+Collective decision-making is likely preferable:
```
jyasskin, line 1507: We could soften the whole recommendation using:

```suggestion
Collective decision-making should be considered:
```

index.html
```diff
@@ -1504,57 +1504,69 @@
...
+    or requests to process data happen very frequently.
+  </li>
+
+  <li>
+    When individuals have systematically less power
+    than the organizations asking them to agree to [=data processing=] ([[?Relational-Turn]]).
+  </li>
+
+  <li>
+    When the [=data processing=] is justified or unjustified at a societal level,
```
npdoty, line 1531:
```suggestion
    When the [=data processing=] is unjust at a societal level even if an individual remains anonymous ([[?Relational-Governance]]).
```

index.html
```diff
@@ -1504,57 +1504,69 @@
...
+    even accounting for individual disagreements ([[?Relational-Governance]]).
```
npdoty, line 1532:
```suggestion
```

darobin (approved):

torgo (approved):
lgtm and reflects consensus of today's call

index.html
jyasskin: In a separate PR. :)

PR #393, opened 2024-01-11
RFC2119 changes

### LINKED ISSUE #371: Be consistent/intentional on "must" vs "should" and RFC 2119 language
Are we intending to use RFC 2119 normative language when we say "must" or "should" in these principles?

If so, we should cite 2119.

In general, it seems like principles are typically "should" language -- they are high-level and involve implementation with many other considerations. But we may want to review which ones say "must" and whether those are absolute.

darobin:
Fixes #371

***
<a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/pull/393.html" title="Last updated on Jan 17, 2024, 5:32 PM UTC (3c4e0bb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/393/da36b9f...3c4e0bb.html" title="Last updated on Jan 17, 2024, 5:32 PM UTC (3c4e0bb)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1183,7 +1183,7 @@ according to their [=users=]' wishes.
 should restrict the [=data=] they transfer to what's either necessary to achieve their users'
 goals or aligns with their users' wishes and interests.</span></div>
 
-<div class="practice" data-audiences="api-designers"><span class="practicelab">Web APIs must be designed to minimize the amount of data that sites need
+<div class="practice" data-audiences="api-designers"><span class="practicelab">Web APIs should be designed to minimize the amount of data that sites need
 to request to carry out their users' goals.
 Web APIs should also provide granularity and user controls over <a>personal
 data</a> that is communicated to sites.</span></div>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1183,7 +1183,7 @@
 should restrict the [=data=] they transfer to what's either necessary to achieve their users'
 goals or aligns with their users' wishes and interests.</span></div>
 
-<div class="practice" data-audiences="api-designers"><span class="practicelab">Web APIs should be designed to minimize the amount of data that sites need
+<div class="practice" data-audiences="api-designers"><span class="practicelab">Web APIs must be designed to minimize the amount of data that sites need
```
jyasskin, line 1186: This one has tradeoffs: letting callers minimize the data increases the size of the API ... so I think it should remain "should".

jyasskin (approved):

index.html
```diff
@@ -2389,6 +2389,20 @@
 
 <section class="appendix" id="bp-summary"></section>
 
+<section class="appendix">
+
+# Conformance
+
+This document does not adhere to strict [[?RFC2129]] terminology because it is primarily of
```
jyasskin, line 2396:
```suggestion
This document does not adhere to strict [[?RFC2119]] terminology because it is primarily of
```

index.html
```diff
@@ -2389,6 +2389,20 @@
...
+an informative nature and does not easily lend itself to constraining a conformance class.
+However, within the formulation of its principles, we have taken care to use "should" to indicate 
+that a principle can be ignored in some rare cases given that there are valid reasons for doing so and
```
npdoty, line 2399: I don't think we mean "can be ignored", but rather that a principle could be overridden in particular cases where there are valid reasons for doing so

jyasskin, line 2399: @npdoty suggested:
```suggestion
that a principle can be overridden in some rare cases given that there are valid reasons for doing so and 
```

PR #437, opened 2024-09-25
Explain how a "3 visits free" model can be compatible with the identity principle.

jyasskin:


jyasskin:
There's no consensus in the task force to merge this.

jyasskin:

PR #438, opened 2024-09-26
Rename "ancillary" to "supporting" and "non-ancillary" to "essential."

jyasskin:
***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/438.html" title="Last updated on Feb 6, 2025, 7:06 PM UTC (386378a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/438/d903f58...jyasskin:386378a.html" title="Last updated on Feb 6, 2025, 7:06 PM UTC (386378a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1290,17 +1290,17 @@ the data exposed, is to autogenerate a new email address for each context.
 
 </aside>
 
-### Ancillary uses
+### Supporting uses
 
 [=Sites=] sometimes use data in ways that aren't needed for the user's immediate
 goals. For example, they might bill advertisers, measure site performance, or
-tell developers about bugs. These uses are known as <dfn data-lt="ancillary
-use">ancillary uses</dfn>.
+tell developers about bugs. These uses are known as <dfn data-lt="supporting
+use">supporting uses</dfn>.
 
-[=Sites=] can get the data they want for [=ancillary uses=] from a variety of places:
+[=Sites=] can get the data they want for [=supporting uses=] from a variety of places:
 
 <dl>
-  <dt><dfn>Non-ancillary APIs</dfn></dt>
+  <dt><dfn>Essential APIs</dfn></dt>
   <dd>
     Web APIs that were designed to support users' immediate goals, like <a
     data-cite="DOM#interface-event">DOM events</a> and <a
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1251,37 +1251,37 @@
 
 </aside>
 
-### Ancillary uses
+### Supporting uses
 
 [=Sites=] sometimes use data in ways that aren't needed for the user's immediate
```
jyasskin, line 1256: Suggested by @wseltzer for a separate PR:
```suggestion
[=Sites=] sometimes use data in ways that aren't directly needed for the user's immediate
```

jyasskin: Sent as https://github.com/w3ctag/privacy-principles/pull/444.

chrisn (commented):
I have a slight concern about changing "ancilliary" to "supporting", as "ancilliary" is somewhat neutral whereas "supporting" may be interpreted as describing these uses as more benign, particularly given the document says the task force doesn't have consensus on how the APIs should be handled.

index.html
```diff
@@ -1251,37 +1251,37 @@
...
+  <dt><dfn>Essential APIs</dfn></dt>
   <dd>
     Web APIs that were designed to support users' immediate goals, like <a
     data-cite="dom#interface-event">DOM events</a> and <a
     data-cite="cssom-view-1#extension-to-the-element-interface">element position
     observers</a>.
   </dd>
 
-  <dt><dfn>Ancillary APIs computed from existing information</dfn></dt>
+  <dt><dfn>Supporting APIs computed from existing information</dfn></dt>
```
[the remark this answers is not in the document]
fantasai: Maybe "auxiliary" would work?

fantasai:
Another approach might be "primary" vs "secondary" instead of "non-ancillary" and "ancillary".

martinthomson (commented):
This is two changes and I'm not a fan of one of the two.

Moving from non-ancillary to essential really does clarify things for me.  It's not just about the negation of the negation, but about the non-optionality involved.  (Though reading some of the ways in which "non-ancillary" was used, I have reservations.  Is the event timing API *really* essential?)

Moving from Ancillary to Supporting makes a value judgment that I don't like.  Curiously, it fits pretty well in some ways, but only by changing who is the focus of the statement.  Consider the "billing advertisers" case.  Labeling that as "ancillary" makes sense for the site visitor when they are the focus of the statement.  Labeling it as "supporting" makes more senses when the website is the implicit focus.

One of the nice things about this document is that it places end users in focus virtually 100% of the time.  This change would go some way to dilute that.  Even if "ancillary" is a bit pompous, I'd want a better word than "supporting" to replace it.

jyasskin:

PR #442, opened 2024-09-28
Say users can say whether pieces of information are sensitive.

jyasskin:
***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/442.html" title="Last updated on Sep 28, 2024, 3:38 AM UTC (c7a41f0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/442/0889270...jyasskin:c7a41f0.html" title="Last updated on Sep 28, 2024, 3:38 AM UTC (c7a41f0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1515,7 +1515,10 @@ For example:
 * [Whether they are using assistive technology.](https://w3ctag.github.io/design-principles/#do-not-expose-use-of-assistive-tech)
 
 A particular piece of information may have different sensitivity for different
-people. People can become vulnerable if sensitive information about them is,
+people. This implies that individual people should be able to adjust whether each
+piece of information about them should be treated as sensitive.
+
+People can become vulnerable if sensitive information about them is,
 or is likely to be, exposed; see [[[#vulnerability]]].
 
 <aside class="example">


index.html
```diff
@@ -1515,7 +1515,10 @@
 * [Whether they are using assistive technology.](https://w3ctag.github.io/design-principles/#do-not-expose-use-of-assistive-tech)
 
 A particular piece of information may have different sensitivity for different
-people. People can become vulnerable if sensitive information about them is,
+people. This implies that individual people should be able to adjust whether each
+piece of information about them should be treated as sensitive.
```
martinthomson, line 1519: This implies privacy labor.  Is that what you really want?

pes10k:
Just following up on our discussion from the call. My 2c is still that no new text is needed, and that the document would be strongest with no changes (in this regard).

However, I said if we were going to make a change, I thought using the font enumeration/fingerprinting trade off as an example block would be a better change than the current PR text. But my preference is still no change is best. 

@dmarti I remember you also had thoughts on this. Can you remind me, what your thoughts and preferences were (PR text vs no changes vs font fingerprinting as an example vs other text)?

dmarti:
@pes10k I think there's a difference between

 * people should be able to share information deliberately
 
  * people should be able to classify information as non-sensitive
  
I would prefer no changes -- if a UA wants to add a sharing control that's a lot more understandable than some kind of sensitive/non-sensitive classifier (that you would have to go back to and re-check every time ML learned to infer sensitive qualities from non-sensitive user data anyway)

pes10k:
okie dokie, if we're at 2-for-2 for no changes, @jyasskin, it might be best to assign someone else. Or if theres no consensus for new text, maybe we just leave the document as is an the council can consider. WDYT?

jyasskin:
I think we should say there's no consensus among the task force to make a chance, and let the council have at it. Thanks for putting your opinions here as input for the council.

jyasskin:

PR #444, opened 2024-10-23
Add that ancillary APIs aren't _directly_ needed for the user's immediate goals.

jyasskin:
***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/444.html" title="Last updated on Oct 23, 2024, 4:48 PM UTC (c804a3e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/444/4337d7b...jyasskin:c804a3e.html" title="Last updated on Oct 23, 2024, 4:48 PM UTC (c804a3e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1309,7 +1309,7 @@ the data exposed, is to autogenerate a new email address for each context.
 
 ### Ancillary uses
 
-[=Sites=] sometimes use data in ways that aren't needed for the user's immediate
+[=Sites=] sometimes use data in ways that aren't directly needed for the user's immediate
 goals. For example, they might bill advertisers, measure site performance, or
 tell developers about bugs. These uses are known as <dfn data-lt="ancillary
 use">ancillary uses</dfn>.


torgo:
This seems redundant since we already talk about "immediate" goals.

martinthomson:
To @torgo's point, I think that the word "directly" is a fine addition, but it might be a better substitute for "immediate".

There are a lot of debates in some groups about indirect benefits and that is where I think that the discussion in this section is addressing that as much as it is about triaging between different goals that a user might have.

Focus on "immediate" says that there is an obvious goal and other goals are secondary.  That carries the risk that your inferences about user motives are wrong.  Maybe someone is doing something because they are aware of the side effects and seek those.

Focus on "direct" implies that those indirect benefits a user might realize from data use (the site can bill advertisers, therefore the site continues to provide whatever service in the future) is key.  I tend to think that this aspect of the discussion is more important in this consideration than questions about user motive and goals, because it separates out the question of intent (which is important) from the question of benefit.

That's a potentially controversial position to take because immediacy is one of the best ways we have of inferring *consent* for use.  So reducing emphasis on that as a means of interpreting user intent weakens any claims a site might have about these ancillary uses.  However, my perspective here is that these questions are tied up in how much you believe consent to be the overriding consideration, as opposed to a focus on effect, specifically end-user benefit.

Both factors are important.  This isn't a simple argument for replacement.  So I have no concern with including both, but I do prefer shifting the emphasis to direct user benefit over indirect.  Similarly, shifting emphasis away from consent (and inference of intent) also seems like a good move.

...that's a lot to write in a PR review for a one word change... ¯\\\_(ツ)\_/¯

csarven:
My reading of the text is that adding "directly" may blur the line between present intent and long-term or indirect benefits. Since the existing distinction is clear, its absence does not create ambiguity.

jyasskin:
Closing now that the [AC review for the changes](https://www.w3.org/wbs/33280/PPrinciplesChanges/results/) is finished, and we didn't get any objections to this wording.

PR #436, opened 2024-09-25
Discuss sites that users pay for with data.

jyasskin:
Fixes #189.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/436.html" title="Last updated on Oct 23, 2024, 4:42 PM UTC (48a21c9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/436/4337d7b...jyasskin:48a21c9.html" title="Last updated on Oct 23, 2024, 4:42 PM UTC (48a21c9)">Diff</a>

Issue #431, opened 2024-09-23
Address the longer Formal Objection


jyasskin:
The objection is in https://lists.w3.org/Archives/Public/public-review-comments/2024Sep/0013.html. The whole thing is quoted here, with [a task list](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/about-task-lists) of PRs that address parts of the objection. Not all of the objection was solvable (which is noted inline), and it's possible not all of these PRs will be accepted.

> We appreciate W3C's initiative in creating the Privacy Principles 
> document aimed at securing user privacy on the web. However, we believe 
> several key aspects of this document require revision to avoid 
> unintended consequences . While the goal of enhancing user privacy is 
> critical, the current draft contains overly broad language, fails to 
> account for the practical realities of content monetization, compromises 
> user choice, and overreaches into regulatory territory. These issues 
> could stifle innovation, compromise the open web, and unfairly impact 
> sectors like advertising, which not only funds the majority of free 
> content but also delivers value to users by making web experiences more 
> relevant and accessible. A careful balance between privacy protections 
> and data availability is essential to ensure that users continue to 
> benefit from personalized and valuable content while their privacy is 
> respected.
> 
> 1. Overly Broad Language and Ambiguity
> The Privacy Principles document is excessively long and complex, making 
> it difficult to understand, implement, and enforce.

- [x] https://github.com/w3ctag/privacy-principles/pull/443 This won't resolve this part of the objection, but it'll improve the situation.

> Principles should be 
> concise and straightforward to avoid contradictions and loopholes. When 
> guidelines are too detailed, they become prone to misinterpretation, 
> which can lead to inconsistent application and enforcement. Clear, short 
> principles ensure that they are widely understood and easily adopted by 
> all stakeholders, fostering more effective compliance across the web 
> ecosystem.
>
> One of the most pressing concerns is the document’s use of overly broad 
> language. Terms like “true intent”, “true preferences” (section 2.12) 

- [x] https://github.com/w3ctag/privacy-principles/pull/432; We had discussed a change to use "... and not to bias the choice.", but the task force only agreed with removing the word "true".

The "true preferences" sentence doesn't need to be reworded because it's not in the principle itself.

> and “enforcement mechanisms” (section 2) are not well defined, leading 
> to potential misinterpretation.

- [x] https://github.com/w3ctag/privacy-principles/pull/433

> Such vagueness allows for subjective 
[...]

PR #451, opened 2024-11-14
Remove the endorsement of exchanging personal data for content.

jyasskin:
This reverts #436 and commit 254a8d0dbf553563c5b9f7c01d4cc606a5ce6851. Some folks on the TAG objected to the change.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/privacy-principles/pull/451.html" title="Last updated on Nov 14, 2024, 6:30 PM UTC (10afc18)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/451/48e1a5f...jyasskin:10afc18.html" title="Last updated on Nov 14, 2024, 6:30 PM UTC (10afc18)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2097,11 +2097,6 @@ allowed to do so without the [=actor=] retaliating, for instance by artificially
 unrelated feature, by decreasing the quality of the service, or by trying to cajole, badger, or
 trick the [=person=] into opting back into the [=processing=].
 
-Note that because data is valuable, some sites offer an informed, free, and fair exchange of
-appropriate user information for original content that the site created or paid for the creation of.
-If a user chooses not to participate in such an exchange, denying them access to the site is not
-retaliation.
-
 ## Support Choosing Which Information to Present {#support-choosing-info}
 
 <div class="practice" data-audiences="user-agents">


martinthomson (approved):

hober (approved):
I'm not sure if simply removing the text is the best approach, but I don't have alternative text to suggest. I can live with this change.

thegreatfatzby:
Hey @jyasskin I haven't been able to participate in this forum as much as I'd like so sorry if this is me wandering in out of my element...is the object to this language a) that it belongs somewhere but just not here (like maybe it belongs in a Private Advertising Tech group but not the Privacy <no modifier> Group) or b) a more foundational objection?

torgo (approved):
yes. re-reviewing this I'm not sure how this got in here in the first place. :/

dmarti:
@torgo This was a modified version of a general suggestion from the formal objection ( https://github.com/w3ctag/privacy-principles/issues/431 https://github.com/w3ctag/privacy-principles/pull/436 ) 

I tried to split out the really problematic case (where the power relationship between the user and the site does not allow for a meaningful choice to share data) from the case that is more like a subscription or paid download, but at a price of zero.

jyasskin:
@thegreatfatzby If I understand it correctly, it's a more foundational objection, wondering whether data access is so abusable that it should never be a valid thing to trade for access. I think we might be able to agree on _something_ on this topic in a future version of this document, and if the objector wants to insist on this, the Council could always agree to reinstate this wording. As you suggest, it might be productive for an advertising group to propose the next iteration of wording, with an argument that the privacy impacts are reasonable.

Seeing general consensus from the TAG, and remembering the sense that the task force would probably not have added this if I hadn't argued in favor of it, I'm going to merge this.

PR #457, opened 2025-02-13
Clarify what we mean by “ancillary”

fantasai:
The word “ancillary” is fairly obscure even for native English speakers, and many dictionaries do not even provide a definition that conveys the intended meaning as understood by those who chose it. These changes make the intended meaning easier for readers to pick up from context, and addresses the FO on this point raised during AC Review.

Relates to https://github.com/w3ctag/privacy-principles/issues/431

***
<a href="https://pr-preview.s3.amazonaws.com/fantasai/privacy-principles/pull/457.html" title="Last updated on Feb 14, 2025, 12:25 AM UTC (36238b4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3ctag/privacy-principles/457/5206db9...fantasai:36238b4.html" title="Last updated on Feb 14, 2025, 12:25 AM UTC (36238b4)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1292,10 +1292,11 @@ the data exposed, is to autogenerate a new email address for each context.
 
 ### Ancillary uses
 
-[=Sites=] sometimes use data in ways that aren't needed for the user's immediate
+[=Sites=] sometimes use data in ways that aren't needed for the user's primary
 goals. For example, they might bill advertisers, measure site performance, or
-tell developers about bugs. These uses are known as <dfn data-lt="ancillary
-use">ancillary uses</dfn>.
+tell developers about bugs. These uses, which provide indirect support for,
+but are not inherent in, the user's primary goals, are known as
+<dfn data-lt="ancillary use">ancillary uses</dfn>.
 
 [=Sites=] can get the data they want for [=ancillary uses=] from a variety of places:
 


index.html
```diff
@@ -1292,10 +1292,11 @@
...
 ### Ancillary uses
 
-[=Sites=] sometimes use data in ways that aren't needed for the user's immediate
+[=Sites=] sometimes use data in ways that aren't needed for the user's primary
 goals. For example, they might bill advertisers, measure site performance, or
-tell developers about bugs. These uses are known as <dfn data-lt="ancillary
-use">ancillary uses</dfn>.
+tell developers about bugs. These uses, which provide indirect support for,
+but are not inherent in, the user's primary goals, are known as
+<dfn data-lt="ancillary use">ancillary uses</dfn>.
```
fantasai, line 1300: FWIW, I debated between “inherent in” and “fundamental to” here. The former seems more accurate to the point, the second reads a little more comfortably to me. I suspect either one could work.

fantasai:
Fwiw, I debated between “inherent in” and “fundamental to”, and settled on the first one because it's slightly more accurate even though it reads a bit less comfortably to me. I suspect either could work.

jyasskin (approved):

rinchen:
The wording changes look good on my end. Thank you.

index.html
```diff
@@ -1292,10 +1292,11 @@
 
 ### Ancillary uses
 
-[=Sites=] sometimes use data in ways that aren't needed for the user's immediate
+[=Sites=] sometimes use data in ways that aren't needed for the user's primary
 goals. For example, they might bill advertisers, measure site performance, or
-tell developers about bugs. These uses are known as <dfn data-lt="ancillary
-use">ancillary uses</dfn>.
+tell developers about bugs. These uses, which provide indirect support for,
```
pes10k, line 1297: I'd prefer

```
These uses might provide indirect support for the user's primary goals, and are known as
<dfn data-lt="ancillary use">ancillary uses</dfn>.
```

I appreciate this might seem a very minor change, but one major concern around "ancillary data" is that the user agent is sharing a bunch of information about the user and user's behavior with site, without anyway of knowing whether its in the user's interest.

fantasai: We could add modifiers like can/might/may/often to "provide" if the intention is that "ancillary" should cover uses that don't support the user's goals at all, even indirectly; but I think the current sentence structure does a better job of actually clarifying what we mean by "ancillary" for two reasons: it connects up the term more directly with the defining phrase, and it contrasts the ancillary uses with essential uses to clarify the relationship between the two categories.

index.html
```diff
@@ -1292,10 +1292,11 @@
...
+but are not inherent in, the user's primary goals, are known as
```
martinthomson, line 1298:
```suggestion

```

index.html
```diff
@@ -1292,10 +1292,11 @@
...
+<dfn data-lt="ancillary use">ancillary uses</dfn>.
```
martinthomson, line 1299: (Apologies for the aborted comment, operator error.)

I agree with @pes10k, but might go further:

```suggestion
tell developers about bugs.

An <dfn>ancillary use</dfn> is any case where [=data processing=]
provides direct benefit to [=actors=]
other than the [=person=] who is the subject of that [=data=].
Ancillary uses of [=personal data=] might benefit to [=people=],
but any benefit is indirect.
For example, being able to charge advertisers provides benefit in that
the site can sustain its business and remain available for future interactions,
but this benefit is primarily realized by business owners.
```

fantasai: Conducting a sales transaction on a website benefits the seller as well as benefiting me. It therefore qualifies as an "ancillary use" of a shopping website under your definition, but I don't think that's what you mean.

martinthomson: I see your point, but would suggest that the addition of the word "only" addresses that concern.

hober (approved):
If @martinthomson's suggested change is also incorporated, this looks good to me.

index.html
fantasai: Ok, well if you're going with this alternate proposal, there are a couple other problems:
1. The paragraph does not connect with the previous paragraph, which is something I tried to ensure with my wording.
2. "Ancillary uses of [=personal data=] might benefit to [=people=]," is ungrammatical and also unclear. What people?

tprieur:
Hi all, This new version looks good to me, thanks !

index.html
martinthomson, line 1297:
```suggestion
tell developers about bugs. These [=ancillary uses=], which provide indirect support for,
```

If you need a connection between paragraphs, this might do the job.

fantasai: You're removing that sentence in your suggested changes, though, so you can't add markup to it.

martinthomson: Oh, that's right, Let me try again.

fantasai (commented):
Heya @martinthomson @hober @pes10k @tprieur ... I wanted to get some clarity on what you all are approving of. For example

- @hober's comment approves of this PR, but only with the addition of Martin's [changes](https://github.com/w3ctag/privacy-principles/pull/457#discussion_r1967032791)... Except Martin's changes basically replace the entire PR with different text so they can't be “also incorporated”. They can only be adopted _instead_ of this PR.
- @martinthomson is trying to address my comment about a poor paragraph transition in [his proposed changes](https://github.com/w3ctag/privacy-principles/pull/457#discussion_r1967032791) by [suggesting a change to my proposed text](https://github.com/w3ctag/privacy-principles/pull/457/files#r1986603662) ... which his proposed changes elide entirely.
- @tprieur approved this PR... but it's unclear what he thinks of Martin's alternative text.

Can we untangle this conversation somehow? CC @jyasskin

martinthomson:
That point about replacing removed text has been addressed with some bridging text in the latest suggestion.  I think that the main question is about @tprieur's disposition toward the original vs. the amendment as both @hober and @pes10k seem to prefer the text I suggested.

index.html
fantasai: It still reads very uncomfortably. Would you consider `s/[=people=]/that [=person=]/`?

And also `s/by business owners/by the site's owners/`.

Because you're mixing up the general and the specific throughout this paragraph and it's awkward.

martinthomson: You know, this might be easier with a new pull request, given how far down the rabbit hole we seem to be.  I've opened https://github.com/w3ctag/privacy-principles/pull/459 where you can make suggestions more directly.

martinthomson:
Just to repeat a comment: I've opened #459 so that we can talk about the alternative text.

jyasskin:
Closing now that the [AC review for the changes](https://www.w3.org/wbs/33280/PPrinciplesChanges/results/) is finished, and we didn't get any objections to this wording.

